PR #5, opened 2023-01-12
4 add interactive configuration api with author

### LINKED ISSUE #4: add interactive configuration api with author

gschennersie:
Version with @susuhahnml as author.


asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
+Copyright (c) 2022 Siemens AG Oesterreich
+SPDX-License-Identifier: MIT
+
+# Documentation of Interactive configurator
```
rtaupe:
```suggestion
# Documentation of Interactive Configurator
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+## API classes
+
+### Knowledge Base
+
+The Knowledge Bases are handled using the `KB` class in [ooasp/kb.py](ooasp/kb.py). This class provides usefully functionality to load, edit, and visualize Knowledge Bases. The string representation of the objects of this class are ASP facts.
+
+### Configuration
+
+The Configurations are handled using the `Config` class in [ooasp/config.py](ooasp/config.py). This class provides usefully functionality to load, edit, and visualize Knowledge Bases.
```
rtaupe:
```suggestion
The Configurations are handled using the `OOASPConfiguration` class in [ooasp/config.py](ooasp/config.py). This class provides useful functionality to load, edit, and visualize Configurations.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
+Copyright (c) 2022 Siemens AG Oesterreich
+SPDX-License-Identifier: MIT
+
+# Documentation of Interactive configurator
+
+## API classes
+
+### Knowledge Base
+
+The Knowledge Bases are handled using the `KB` class in [ooasp/kb.py](ooasp/kb.py). This class provides usefully functionality to load, edit, and visualize Knowledge Bases. The string representation of the objects of this class are ASP facts.
```
rtaupe:
```suggestion
The Knowledge Bases are handled using the `OOASPKnowledgeBase` class in [ooasp/kb.py](ooasp/kb.py). This class provides useful functionality to load, edit, and visualize Knowledge Bases. The string representation of the objects of this class are ASP facts.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Configuration
+
+The Configurations are handled using the `Config` class in [ooasp/config.py](ooasp/config.py). This class provides usefully functionality to load, edit, and visualize Knowledge Bases.
+Furthermore, it can serve as container for partial configurations, complete configurations, configuration checks and available options for a configuration.  The string representation of the objects of this class are ASP facts.
```
rtaupe:
```suggestion
Furthermore, it can serve as container for partial configurations, complete configurations, configuration checks and available options for a configuration. The string representation of the objects of this class are ASP facts.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+### Interactive Configurator
+
+The interactive Configurator API provides all functionality to create a configuration interactively. This is done by constructing an `InteractiveConfigurator` class from [ooasp/interactive.py](ooasp/interactive.py). The process is done in a [multi shot](#multi-shot-approach) way.
```
rtaupe:
```suggestion
The interactive Configurator API provides all functionality to create a configuration interactively. This is done by calling the methods provided by the `InteractiveConfigurator` class from [ooasp/interactive.py](ooasp/interactive.py). The process is done in a [multi-shot](#multi-shot-approach) way.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+#### Tasks
+
+The following tasks correspond to the only points of the interactive process where solving is performed. They are done for the current domain size and based on a (possibly empty) partial configuration `C`. 
+
+1. *Complete:* Getting a complete configuration extending `C`.
+2. *Check:* Checking `C` for errors.
+3. *Options:* Obtaining all possible leafs for the objects, values for the attributes and associations to complete `C`.
```
rtaupe:
```suggestion
3. *Options:* Obtaining all possible classes for the objects, values for the attributes, and associations to complete `C`. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+----
+## Encodings
+
+All encodings are put together using a single file:
+
+- [ooasp/encodings/ooasp.lp](ooasp/encodings/ooasp.lp)
+
+This file also introduces **external atoms** to differentiate the task that will be performed without the need for re-grounding. The truth value of those atoms will be defined eternally inside the [Interactive Configurator](#interactive-configurator).
```
rtaupe:
```suggestion
This file also introduces **external atoms** to differentiate the task that will be performed without the need for re-grounding. The truth value of those atoms will be defined externally inside the [Interactive Configurator](#interactive-configurator).
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+**Externals**
+
+- **`guess`**  The [guessing](#guess) part is active. This external is used instead of the old predicate `ooasp_configure`.
```
rtaupe:
```suggestion
- **`guess`**  The [guessing](#guess) part is active. This external is used instead of the old predicate [`ooasp_configure`](../ooasp/ooasp_config.lp).
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+  
+Task |  Value |
+-----|---------
+*Complete*     |   `true`     |
+*Check*    |   `false`     |
+*Options*    |   `true`     |
```
rtaupe:
```suggestion
	Value of this external for the different tasks:
	Task |  Value |
	-----|---------
	*Complete*     |   `true`     |
	*Check*    |   `false`     |
	*Options*    |   `true`     |
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`check`**  The check part is active for [complete constraints](#complete-constraints). This refers only to the integrity constrains, however `ooasp_cv` atoms are always derived, regardless of the external values.
+  
+Task |  Value |
+-----|---------
+*Complete*     |   `true`     |
+*Check*    |   `false`     |
+*Options*    |   `true`     |
```
rtaupe:
```suggestion
	Value of this external for the different tasks:
	Task |  Value |
	-----|---------
	*Complete*     |   `true`     |
	*Check*    |   `false`     |
	*Options*    |   `true`     |
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+Although this might seam contra-intuitive, this external should be false for the *Checking* task in order to have the `ooasp_cv` atoms as part of a stable model.
+
+- **`check_partial_cv`**  The check part is active for [partial constraints](#partial-constraints). This refers only to the integrity constrains, however `ooasp_cv` atoms are always derived, regardless of the external values.
+  
+Task |  Value |
+-----|---------
+*Complete*     |   `true`     |
+*Check*    |   `false`     |
+*Options*    |   `false`     |
```
rtaupe:
```suggestion
	Value of this external for the different tasks:
	Task |  Value |
	-----|---------
	*Complete*     |   `true`     |
	*Check*    |   `false`     |
	*Options*    |   `false`     |
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+  
+Task |  Value |
+-----|---------
+*Complete*     |   `true`     |
+*Check*    |   `false`     |
+*Options*    |   `true`     |
+
+Although this might seam contra-intuitive, this external should be false for the *Checking* task in order to have the `ooasp_cv` atoms as part of a stable model.
+
+- **`check_partial_cv`**  The check part is active for [partial constraints](#partial-constraints). This refers only to the integrity constrains, however `ooasp_cv` atoms are always derived, regardless of the external values.
```
rtaupe:
```suggestion
- **`check_partial_cv`**  The check part is active for [partial constraints](#partial-constraints). This refers only to the integrity constraints, however `ooasp_cv` atoms are always derived, regardless of the external values.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`check`**  The check part is active for [complete constraints](#complete-constraints). This refers only to the integrity constrains, however `ooasp_cv` atoms are always derived, regardless of the external values.
+  
+Task |  Value |
+-----|---------
+*Complete*     |   `true`     |
+*Check*    |   `false`     |
+*Options*    |   `true`     |
+
+Although this might seam contra-intuitive, this external should be false for the *Checking* task in order to have the `ooasp_cv` atoms as part of a stable model.
```
rtaupe:
```suggestion
Although this might seem contra-intuitive, this external should be false for the *Checking* task in order to have the `ooasp_cv` atoms as part of a stable model.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+Auxiliary predicates used for both Knowledge Bases and Configurations can be found in the following files. Note that auxiliary predicates for the configuration are **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
+
+- [ooasp/encodings/ooasp_aux_kb.lp](ooasp/encodings/ooasp_aux_kb.lp)
+- [ooasp/encodings/ooasp_aux_config.lp](ooasp/encodings/ooasp_aux_config.lp)
+
+
+### Guess
+
+The guessing section contains all choices to decide the leaf class of an object, the associations, and the values. Note that the guessing is **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
```
rtaupe:
```suggestion
The guessing section contains all choices to decide the class of an object, the associations, and the values. Note that the guessing is **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+- [ooasp/encodings/ooasp_guess.lp](ooasp/encodings/ooasp_guess.lp)
+
+
+### Constraints
+
+Constraints will be **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
+
+- [ooasp/encodings/ooasp_check.lp](ooasp/encodings/ooasp_check.lp)
+  
+Constraints are defined using predicate:
```
rtaupe:
```suggestion
Constraints are defined using the `ooasp_cv/5` predicate:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+**`ooasp_cv(CONFIG,CV_NAME,OBJECT,STR,ARGS)`**
+- `CONFIG`: The name of the configuration
+- `CV_NAME`: The name of the constraint. Should be unique, as it is used for defining partial constraints
+- `OBJECT`: The identifier of the object to which this constraint violation refers to. This is used to better visualize constraints and to allow incremental grounding.
+- `STR`: A string describing the constraint. It can use place holders `{}`, which are filled with the arguments in `ARGS`.
+- `ARGS`: A tuple with the arguments to format `STR`. Note that tuples with a single element need to be written as `(ARG,)`
+
+
+
+**Domain specific constrains** are defined by the user in a different file, which can be provided when the [Interactive Configurator](#interactive-configurator) is created. Note that these will also be **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
```
rtaupe:
```suggestion
**Domain specific constraints** are defined by the user in a different file, which can be provided when the [Interactive Configurator](#interactive-configurator) is created. Note that these will also be **grounded incrementally** (see [multi-shot](#multi-shot-approach)).
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+#### Complete constraints
+
+Those constrains that can not longer be corrected by adding new values and associations. These constraints will make a configuration invalid and thus, are not provided as options for the user. By default all constraints are considered of this type, unless stated otherwise (see bellow [Partial constrains](#partial-constraints)).
```
rtaupe:
```suggestion
Those constrains that can not longer be corrected by adding new values and associations. These constraints will make a configuration invalid and thus, are not provided as options for the user. By default all constraints are considered of this type, unless stated otherwise (see below [Partial constraints](#partial-constraints)).
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+*For example: Upper bound constraints or assigning an invalid value*
```
rtaupe:
```suggestion
*For example: Upper-bound constraints or assigning an invalid value.*
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+#### Partial constraints
+
+Those constraints that may be violated in the current (partial) configuration but can become satisfied at a later stage of the configuration process. 
+
+*For example: Lower bound constraints, missing value for an attribute.*
```
rtaupe:
```suggestion
*For example: Lower-bound constraints, missing value for an attribute.*
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+Partial constraints are ignored when using brave reasoning in order to get all possible options for values and associations. They are only checked when the external `ooasp_partial_cv` is true. Nonetheless, the constraint violation atoms `ooasp_cv` for this constrains will still appear when checking the configuration for errors.
```
rtaupe:
```suggestion
Partial constraints are ignored when using brave reasoning in order to get all possible options for values and associations. They are only checked when the external `ooasp_partial_cv` is true. Nonetheless, the constraint violation atoms `ooasp_cv` for these constraints will still appear when checking the configuration for errors. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+### User input
+
+The user input corresponds to the current partial configuration `C` that is being constructed. This is added into the encoding using externals and it is **grounded incrementally** (See [multi-shot](#multi-shot-approach)).
+
+- [ooasp/encodings/ooasp_user_input.lp](ooasp/encodings/ooasp_user_input.lp)
+
+**Externals**
+
+- **`user(ooasp_isa_leaf(CONFIG,LEAFCLASS,ID))`**
+  The user assigned a leaf class to an object `ID`
```
rtaupe:
```suggestion
  The user assigned leaf class `LEAFCLASS` to object `ID`
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`user(ooasp_associated(CONFIG,ASSOC,ID1,ID2))`**
+  The user associated `ID1` and `ID2`
+
+- **`user(ooasp_attribute_value(CONFIG,N,ID,VALUE))`**
+  The user selected a value for object `ID`
```
rtaupe:
```suggestion
  The user selected value `VALUE` for attribute `N` of object `ID`
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+The truth value of this externals is defined by the [Interactive Configurator](#interactive-configurator) based on the current (partial) configuration that is being constructed using an object of class [Config](#configuration).
```
rtaupe:
```suggestion
The truth value of these externals is defined by the [Interactive Configurator](#interactive-configurator) based on the current (partial) configuration that is being constructed using an object of class [OOASPConfiguration](#configuration). 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`user(ooasp_attribute_value(CONFIG,N,ID,VALUE))`**
+  The user selected a value for object `ID`
+
+The truth value of this externals is defined by the [Interactive Configurator](#interactive-configurator) based on the current (partial) configuration that is being constructed using an object of class [Config](#configuration). 
+
+
+----
+
+## Multi shot approach
```
rtaupe:
```suggestion
## Multi-shot approach
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+This approach intends to tackle the bottle neck issue of re-grounding rules multiple times after each change.
```
rtaupe:
```suggestion
This approach intends to tackle the bottleneck issue of re-grounding rules multiple times after each change. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+----
+
+## Multi shot approach
+
+This approach intends to tackle the bottle neck issue of re-grounding rules multiple times after each change. 
+
+
+
+The rules are grounded for every new object that is introduced in an incremental way. All encodings that starting with the directive `#program domain(new_id).` will be grounded when a new object is introduced into the configuration. This new object will have the id `new_id`.
```
rtaupe:
```suggestion
The rules are grounded for every new object that is introduced in an incremental way. The subprogram starting with the directive `#program domain(new_id).`, which is split among several encoding files, will be grounded when a new object is introduced into the configuration. This new object will have the id `new_id`.
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -0,0 +1,148 @@
...
+%=======================================
+% Values
+%=======================================
+
+%--- (USER INPUT) Attribute value must belong to instance class
+
+ooasp_cv(CONFIG,no_instance_for_attribute,new_id,"Attribute {} not of selected class",(ATTR,)) :- 
+	ooasp_attribute(V,C1,ATTR,T),
+	ooasp_attribute_value(CONFIG,ATTR,new_id,VALUE),
+	not ooasp_isa(CONFIG,C1,new_id).
```
rtaupe, line 98: This constraint violation may be wrongly produced if attributes of the same name appear in two different classes.
Attention: This code snippet also appears in `DOC.md`!

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+	ooasp_attribute(V,C1,ATTR,T),
+	ooasp_attribute_value(CONFIG,ATTR,new_id,VALUE),
+	not ooasp_isa(CONFIG,C1,new_id).
+```
+
+This rule will be grounded for a `new_id`, since this value appears in the head we know it hasn't been grounded before.
+
+##### Example (Multiple positions)
+
+In this example we want to generate the constraint violations when an association is not of the right class, using on predicate `ooasp_associated(CONFIG,ASSOC,ID1,ID2)`. We must notice that our `new_id` could be either of the two values `ID1` or `ID2`, this means that we need two rules considering both cases.
```
rtaupe:
```suggestion
In this example we want to generate the constraint violations when an association is not of the right class, using predicate `ooasp_associated(CONFIG,ASSOC,ID1,ID2)`. We must notice that our `new_id` could be either of the two values `ID1` or `ID2`, this means that we need two rules considering both cases.
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -0,0 +1,148 @@
...
+	ooasp_configuration(V,CONFIG),
+	ooasp_associated(CONFIG,ASSOC,new_id,_),
+	ooasp_assoc(V,ASSOC,C1,C1MIN,C1MAX,C2,C2MIN,C2MAX),
+	not ooasp_isa(CONFIG,C1,new_id).
+
+ooasp_cv(CONFIG,wrongtypeinassoc,new_id,"Associated by {} but is not of class {}",(ASSOC,C2)) :-
+	ooasp_configuration(V,CONFIG),
+	ooasp_associated(CONFIG,ASSOC,_,new_id),
+	ooasp_assoc(V,ASSOC,C1,C1MIN,C1MAX,C2,C2MIN,C2MAX),
+	not ooasp_isa(CONFIG,C2,new_id).
```
rtaupe, line 86: I think we have here a similar problem as with the cv `no_instance_for_attribute`. Should we be supporting attributes and association that appear with the same name in different classes?
Attention: This code snippet also appears in `DOC.md`!

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+	ooasp_configuration(V,CONFIG),
+	ooasp_associated(CONFIG,ASSOC,_,new_id),
+	ooasp_assoc(V,ASSOC,C1,C1MIN,C1MAX,C2,C2MIN,C2MAX),
+	not ooasp_isa(CONFIG,C2,new_id).
+```
+
+
+### Accumulative values
+
+There are some rules that depend on atoms that are computed an an accumulative way. One of this values is the **arity**.
```
rtaupe:
```suggestion
There are some rules that depend on atoms that are computed in an accumulative way. One of this values is the **arity**. 
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
rtaupe: issue reported as #8

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
rtaupe: issue reported as #8

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+There are some rules that depend on atoms that are computed an an accumulative way. One of this values is the **arity**. 
+
+Consider the following rule:
+
+```prolog
+arity(new_id,ASSOC,ARITY):- ARITY = #count{ID2:ooasp_assoc(V,ASSOC,new_id,_,_,ID2,_,_)}.
+```
+
+This rule will gather the number of objects that `new_id` is associated to in association `ASSOC`. While this is correct at the moment you ground `new_id` this value will not longer be correct when new identifiers are grounded. This is because they are were not part of the aggregate `#count` calculated before. Therefore, associations of `new_id` with any `ID>new_id` will not be counted.
```
rtaupe:
```suggestion
This rule will gather the number of objects that `new_id` is associated to in association `ASSOC`. While this is correct at the moment you ground `new_id` this value will no longer be correct when new identifiers are grounded. This is because they were not part of the aggregate `#count` calculated before. Therefore, associations of `new_id` with any `ID>new_id` will not be counted. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+ooasp_arity(CONFIG, ASSOC, 2, new_id, ARITY, new_id) :-
+	ooasp_configuration(V,CONFIG),
+	ooasp_assoc(V,ASSOC,C1,C1MIN,C1MAX,C2,C2MIN,C2MAX),
+	ooasp_isa(CONFIG,C2,new_id),
+	ARITY = #count { ID1:ooasp_associated(CONFIG,ASSOC,ID1,new_id) }.
+```
+
+These rules compute the arity as shown before for the two possible positions of the new ID. The first one for position `1` and the second one for position `2`. Unlike in our previous rule, here, we also include the `new_id` as the last argument of `ooasp_arity`. We do so to identify that this value `ARITY` for the arity is only valid when `new_id` was grounded. This value can be seen as a step.
+
+Next, we include any new associations that this `new_id` might have with previously grounded ids and update their corresponding arity. This is done in the following rules:
```
rtaupe:
```suggestion
Next, we include any new associations that this `new_id` might have with previously grounded IDs and update their corresponding arity. This is done in the following rules:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+ooasp_arity(CONFIG, ASSOC, 1, ID1, ARITY, new_id) :- 
+	ooasp_arity(CONFIG, ASSOC, 1, ID1, ARITY, new_id-1),
+	not ooasp_associated(CONFIG,ASSOC, ID1, new_id).
+
+ooasp_arity(CONFIG, ASSOC, 2, ID2, ARITY, new_id) :- 
+	ooasp_arity(CONFIG, ASSOC, 2, ID2, ARITY, new_id-1),
+	not ooasp_associated(CONFIG,ASSOC, new_id, ID2).
+```
+
+These rules update the values like a counter without using `#count`. Note that the last argument of the `ooasp_arity` is `new_id`, serving two proposes: first, to identify that this is a valid arity for that moment, and second to ensure that this is a new head that hasn't been grounded before.
```
rtaupe:
```suggestion
These rules update the values like a counter without using `#count`. Note that the last argument of the `ooasp_arity` is `new_id`, serving two purposes: first, to identify that this is a valid arity for that moment, and second to ensure that this is a new head that hasn't been grounded before.
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+
+#### Using accumulative values
+
+If we want to use this accumulative values in another rule there is one more detail to consider.
```
rtaupe:
```suggestion
If we want to use these accumulative values in another rule there is one more detail to consider. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+These rules update the values like a counter without using `#count`. Note that the last argument of the `ooasp_arity` is `new_id`, serving two proposes: first, to identify that this is a valid arity for that moment, and second to ensure that this is a new head that hasn't been grounded before.
+
+
+#### Using accumulative values
+
+If we want to use this accumulative values in another rule there is one more detail to consider. 
+
+**Example**
+
+Image the rule:
```
rtaupe:
```suggestion
Imagine the rule:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+```prolog
+ooasp_cv(CONFIG,wrongass,ID,"Wrong for {}",(ID,)) :- 
+    ooasp_arity(CONFIG,assoc1,1,ID,ARITY,STEP), ARITY!=4.
+```
+
+This rule is saying that for association `assoc1` any the object `ID` in position `1` should have exactly `4` objects associated. We first notice that this is not using the `new_id`. We can include this by doing:
```
rtaupe:
```suggestion
This rule is saying that for association `assoc1` the object `ID` in position `1` should have exactly `4` objects associated. We first notice that this is not using the `new_id`. We can include this by doing:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+```
+
+Notice that we want to ground this rule for the `new_id` this means that `STEP` will be `new_id`. 
+
+```prolog
+ooasp_cv(CONFIG,wrongass,new_id,"Wrong for {}",(new_id,)) :- 
+    ooasp_arity(CONFIG,assoc1,1,new_id,ARITY,new_id), ARITY!=4, active(new_id).
+```
+
+Since we updated the values of `ooasp_arity` of all previous ids as well, we also need to ground the corresponding rules for those, yielding the rule:
```
rtaupe:
```suggestion
Since we updated the values of `ooasp_arity` of all previous IDs as well, we also need to ground the corresponding rules for those, yielding the rule:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+Where the following fact must also included:
+
+```prolog
+ooasp_partial_cv(wrongass_lower).
+```
+
+#### Efficiency
+
+Even though the explanation above corresponds to a correct solution, it is not efficient. Generally speaking, using the computed value of an aggregate such as `#count` in the head of the rule accounts for additional grounding. Ideally, this value should be used only in the body of the rule. This observations leads us to a more efficient solution, where the aggregate is used directly in the constraint.
```
rtaupe:
```suggestion
Even though the explanation above corresponds to a correct solution, it is not efficient. Generally speaking, using the computed value of an aggregate such as `#count` in the head of the rule accounts for additional grounding. Ideally, this value should be used only in the body of the rule. This observation leads us to a more efficient solution, where the aggregate is used directly in the constraint. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+	ooasp_configuration(V,CONFIG),
+	ooasp_assoc_limit(V,ASSOC,max,OPT,C,CMAX),
+	ooasp_isa(CONFIG,C,ID1),
+	#count { ID2:ooasp_associated_general(CONFIG,ASSOC,OPT,ID1,ID2) } > CMAX,
+	active(new_id).
+```
+
+Notice that, as before, this rule must consider the `active(new_id)` to define in which step the constraint is violated. The count, however, does not depend on the `new_id`. 
+
+We use here two axillary predicates to avoid having to repeat the rule:
```
rtaupe:
```suggestion
We use here two auxiliary predicates to avoid having to repeat the rule:
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`ooasp_assoc_limit(V,ASSOC,max,OPT,C,CMAX)`**
+  The association `ASSOC` has a maximum limit of `MAX` when an object of class `C` is of class type `OPT`, where `OPT` is `1` or `2` .
+  *Example: `ooasp_assoc_limit(V,ass,min,1,racks,10)`
+	The association `ass` must be from a `racks` element as class `C1` to at least `10` elements*
```
rtaupe:
```suggestion
- **`ooasp_assoc_limit(V,ASSOC,max,POS,C,CMAX)`**
  The association `ASSOC` has a maximum limit of `CMAX` for objects of class `C` at `POS`, where `POS` is `1` or `2` and refers to one end of the association (1: left end, 2: right end).
  *Example: `ooasp_assoc_limit(V,ass,min,1,racks,10)`:
	In the association `ass`, each `racks` object must be associated to at least `10` elements*
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+- **`ooasp_associated_general(CONFIG,ASSOC,OPT,ID1,ID2)`**
+  Object `ID1` appearing in position `OPT` of association `ASSOC` is associated to object `ID2`
+  *Example: `ooasp_associated_general(c1,ass,2,23,10)` object `23` is the class `C2` of association `ass` with object `10`*
```
rtaupe:
```suggestion
- **`ooasp_associated_general(CONFIG,ASSOC,POS,ID1,ID2)`**
  Object `ID1` appearing in position `POS` of association `ASSOC` is associated to object `ID2`
  *Example: `ooasp_associated_general(c1,ass,2,23,10)`: object `23` of the second class of association `ass` is associated with object `10`*
```

asp_interactive_configuration/DOC.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Incremental solving 
+
+This multi shot encoding can directly be used for incremental solving. With this idea we can ground for id `1` and try to find a configuration using Task 1 *Complete*, if there is no complete configuration (returns UNSAT) with this many object, we ground for the next object `2` and repeat the process until we have a satisfiable answer.
```
rtaupe:
```suggestion
This multi-shot encoding can directly be used for incremental solving. With this idea we can ground for id `1` and try to find a configuration using Task 1 *Complete*, if there is no complete configuration (returns UNSAT) with this number of objects, we ground for the next object `2` and repeat the process until we have a satisfiable answer. 
```

asp_interactive_configuration/README.md
```diff
@@ -0,0 +1,23 @@
+Copyright (c) 2022 Siemens AG Oesterreich
+SPDX-License-Identifier: MIT
+
+# Interactive Configuration with ASP
+
+
+Prototype for interactive configuration using ASP initially developed by Susana Hahn (Potassco Solutions) as part of a collaboration between Potassco Solutions and Siemens T.
```
rtaupe:
```suggestion
Prototype for interactive configuration using ASP initially developed by Susana Hahn (Potassco Solutions) as part of a collaboration between Potassco Solutions and Siemens.
```

asp_interactive_configuration/README.md
```diff
@@ -0,0 +1,23 @@
...
+```
+poetry install
+```
+
+
+## Usage
+
+The files corresponding to the current version are in [`ooasp`](ooasp).
+
+The package usage and new features are showcased in the jupyter notebooks inside the folder [`usage`](usage)
```
rtaupe:
```suggestion
The package usage and new features are showcased in the jupyter notebooks inside the folder [`usage`](usage).
```

asp_interactive_configuration/examples/metro/constraints.lp
```diff
@@ -0,0 +1,167 @@
+% Copyright (c) 2022 Siemens AG Oesterreich
```
rtaupe, line 1: I can't seem to add a comment to the `.png` files, so I'm asking here: What is the difference between `metro_v1.png` and `v1.png`? They look rather similar.

asp_interactive_configuration/examples/metro/constraints.lp
```diff
@@ -0,0 +1,167 @@
...
+
+% currently only one wagon allowed
+
+ooasp_cv(C,only_one_wagon,W1,"Only one wagon allowed {} {}",(new_id,W2)):-
+    active(new_id),
+    ooasp_isa(C,wagon,W1),
+    ooasp_isa(C,wagon,W2),
+    W1!=W2.
+
+% standingroom > 0 -> count(handrail)>0
```
rtaupe, line 22: According to the name of the following two CVs, `->` should probably be replaced by `<->` or `iff`?

asp_interactive_configuration/examples/metro/constraints.lp
```diff
@@ -0,0 +1,167 @@
...
+ooasp_cv(CONFIG,new_id,special_seats_color_mustbe_red,"special_seats_color_mustbe_red {}",(new_id,S)):-
+    active(new_id),
+    ooasp_attribute_value(CONFIG,seat_type,S,TYPE),
+    TYPE=special,
+    ooasp_attribute_value(CONFIG,seat_color,S,COLOR),
+    COLOR!=red.
+
+%*
+deactivated because of performance
+ooas_cv(C,nrpassengers_neq_nrseats_plus_standingroom,W,"{} nr_passengers {} = nr_seats {} = standingroom {}",(new_id,W,NRPASSENGERS,NRSEATS,STANDINGROOM)):-
```
rtaupe:
```suggestion
ooasp_cv(C,nrpassengers_neq_nrseats_plus_standingroom,W,"{} nr_passengers {} = nr_seats {} = standingroom {}",(new_id,W,NRPASSENGERS,NRSEATS,STANDINGROOM)):-
```

asp_interactive_configuration/examples/metro/constraints.lp
```diff
@@ -0,0 +1,167 @@
...
+    active(new_id),
+    ooasp_attribute_value(C,nr_seats,W,NRSEATS),
+    ooasp_attribute_value(C,standing_room,W,STANDINGROOM),    
+    ooasp_attribute_value(C,nr_passengers,W,NRPASSENGERS),    
+    NRPASSENGERS != NRSEATS + STANDINGROOM.
+*%
+
+%
+%* all equal type except special
```
rtaupe, line 100: This CV is commented out. What is the difference to the similar constraints above and why was it commented out?

asp_interactive_configuration/examples/metro/constraints.lp
```diff
@@ -0,0 +1,167 @@
...
+%    ooasp_attribute_value(C,nr_passengers,W,NRPASSENGERS),
+%    ooasp_attribute_value(C,standing_room,W,STANDINGROOM),
+%    NRSEATS = NRPASSENGERS - STANDINGROOM.
+
+ooasp_attribute_value(C,standing_room,W,STANDINGROOM):-
+    ooasp_attribute_value(C,nr_passengers,W,NRPASSENGERS),
+    ooasp_attribute_value(C,nr_seats,W,NRSEATS),
+    STANDINGROOM = NRPASSENGERS - NRSEATS.
+
+*%
```
rtaupe, line 167: Please think again if all these commented-out CVs shall really stay in the file, and if yes, provide a bit of explanation.

asp_interactive_configuration/examples/metro/kb.lp
```diff
@@ -0,0 +1,51 @@
...
+ooasp_attribute(metro_v1,seat,seat_color,str).
+ooasp_attribute_enum(metro_v1,seat,seat_color,blue).
+ooasp_attribute_enum(metro_v1,seat,seat_color,red).
+ooasp_attribute_enum(metro_v1,seat,seat_color,white).
+
+ooasp_attribute(metro_v1,handrail,handrail_type,str).
+ooasp_attribute_enum(metro_v1,handrail,handrail_type,standard).
+ooasp_attribute_enum(metro_v1,handrail,handrail_type,premium).
+
+% constraints
```
rtaupe:
```suggestion
```

asp_interactive_configuration/examples/racks/constraints.lp
```diff
@@ -0,0 +1,240 @@
...
+ooasp_partial_cv(customlowerbound).
+ooasp_partial_cv(racksinglelowerbound).
+ooasp_partial_cv(rackdoublelowerbound).
+ooasp_partial_cv(module_requires_element).
+ooasp_partial_cv(moduleII_requires_moduleV).
+
+#program domain(new_id).
+
+
+ooasp_cv(C,assoc1_constraint1,new_id,"Asssociated by {} to {} which is not of type {}",(new_id,ASSOC,ID2,C2)) :-
```
rtaupe, line 22: These CV texts do not make much sense to me. Also, there are 4 variables but only 3 placeholders!?

asp_interactive_configuration/examples/racks/constraints.lp
```diff
@@ -0,0 +1,240 @@
...
+    P<1.
+
+ooasp_cv(C,frame_must_be_within_range, F, "Frame position is outside of range {} ",(F,new_id)):-
+    ooasp_isa(C,rackDouble,new_id),
+    ooasp_associated(C,rack_frames,new_id,F),
+    ooasp_attribute_value(C,frame_position,F,P),
+    P>8.
+
+% moduleII requires moduleV
+% TODO: Is using active(new_id) enough?
```
rtaupe, line 163: What exactly does this TODO mean?

asp_interactive_configuration/examples/racks/constraints.lp
```diff
@@ -0,0 +1,240 @@
...
+
+ooasp_cv(C,moduleII_requires_moduleV,F,"ModuleII requires moduleV in frame {} new_id {} {} {}",(F,new_id,NROFMODULEII,NROFMODULEV)):-
+    active(new_id),
+    ooasp_isa(C,frame,F),
+    NROFMODULEII = #count { M:ooasp_associated(C,frame_modules,F,M),ooasp_isa(C,moduleII,M) },
+    NROFMODULEV = #count { M:ooasp_associated(C,frame_modules,F,M),ooasp_isa(C,moduleV,M) },
+    NROFMODULEII = 0,
+    NROFMODULEV > 0.
+
+% new frame
```
rtaupe, line 180: Why are the remaining CVs commented out?

asp_interactive_configuration/examples/racks/kb.lp
```diff
@@ -0,0 +1,89 @@
...
+% ooasp_cv(C,module_may_not_have_element(C,ID)):-ooasp_is_associated2_cardinality(C,element_modules,ID,N), N>0, ooasp_isa(C,moduleV,ID).
+
+% ooasp_cv(C,frame_position_must_be_unique(C,F1,F2,P)):-
+%     ooasp_associated(C,rack_frames,R,F1),
+%     ooasp_associated(C,rack_frames,R,F2),
+%     F1!=F2, 
+%     ooasp_attribute_value(C,frame_position,F1,P),
+%     ooasp_attribute_value(C,frame_position,F2,P).
+
+
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+    def unifiers_list(self)->List:
+        """
+        The list of all unifiers classes
+        """
+        return self.UNIFIERS.__dict__.values()
+    
+    @property
+    def domain_size(self)->int:
+        """
+        The domain size, it is computing by couting the nomber of objects in the factbase.
```
rtaupe:
```suggestion
        The domain size, it is computed by counting the number of objects in the fact base.
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        return self.fb.query(self.UNIFIERS.Domain).select(self.UNIFIERS.Domain.object_id).count()
+
+    @property
+    def has_cv(self)->bool:
+        """
+        True if the configuration has any constraint violations, false otherwise
+        """
+        return len(self.constraint_violations) != 0
```
rtaupe, line 152: Isn't `self.constraint_violations` a method and should thus be called with parentheses?

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        Removes the given facts from the configuration factbase
+        """
+        for f in facts:
+            self.fb.remove(f)
+
+    def add_domain(self, class_name:str, object_id:int)->Predicate:
+        """
+        Adds a new domain fact `ooasp_domain`
+            Paramters:
```
rtaupe:
```suggestion
            Parameters:
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+                The added fact
+        """
+        fact = self.UNIFIERS.Domain(class_name=class_name,object_id=object_id)
+        self.fb.add(fact)
+        return fact
+    
+    def add_leaf(self,object_id:int, class_name:str)->Predicate:
+        """
+        Adds a new leaf predicate to the factbase
+            Paramters:
```
rtaupe:
```suggestion
            Parameters:
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+                object_id: The identifier for the object
+                class_name: The class name
+            Returns:
+                The added fact
+            Throws:
+                Exception if the class name is not a leaf class
```
rtaupe:
```suggestion
                Exception if the class of the given name is not a leaf class
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+            raise RuntimeError(f"{class_name} is not a leaf class")
+
+        fact = self.UNIFIERS.Leaf(class_name=class_name,object_id=object_id)
+        self.fb.add(fact)
+        return fact
+
+    def add_value(self, object_id:int, attr_name:str, attr_value)->Predicate:
+        """
+        Adds a new attribute value predicate to the factbase
+            Paramters:
```
rtaupe:
```suggestion
            Parameters:
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+                object_id: The identifier for the object
+                attr_name: The name od the attribute
```
rtaupe:
```suggestion
                attr_name: The name of the attribute
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+                attr_value: The value od the attribute
```
rtaupe:
```suggestion
                attr_value: The value of the attribute
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+        fact = self.UNIFIERS.Association(assoc_name=assoc_name,
+            object_id1=object_id1,
+            object_id2=object_id2)
+        self.fb.add(fact)
+        return fact
+
+    def remove_leaf(self, object_id)->List[Predicate]:
+        """
+        Removes any leaf predicates from the factbase associated to the object id
+            Paramters:
```
rtaupe:
```suggestion
            Parameters:
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        Leaf = self.UNIFIERS.Leaf
+        q = self.fb.query(Leaf).where(Leaf.object_id == object_id)
+        leafs = list(q.all())
+        self._remove_facts(leafs)
+        return leafs
+
+    def remove_value(self, object_id:id, attr_name:str)->List[Predicate]:
+        """
+        Removes any exisitng values for the given attribute of an object in the factbase
```
rtaupe:
```suggestion
        Removes any existing values for the given attribute of an object in the fact base
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+            Paramters:
```
rtaupe:
```suggestion
            Parameters:
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+                object_id: The identifier for the object
+                attr_name: The name od the attribute
```
rtaupe:
```suggestion
                attr_name: The name of the attribute
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+        AttributeValue = self.UNIFIERS.AttributeValue
+        q = self.fb.query(AttributeValue)
+        q = q.where(((AttributeValue.object_id == object_id) & (AttributeValue.attr_name == attr_name)))
+        values = list(q.all())
+        self._remove_facts(values)
+        return values
+
+    def remove_association(self, assoc_name:str, object_id1:int, object_id2:int)->List[Predicate]:
+        """
+        Removes the the association from the factbase
```
rtaupe:
```suggestion
        Removes the association from the fact base
```

asp_interactive_configuration/ooasp/encodings/ooasp_arity.lp
```diff
@@ -0,0 +1,46 @@
+% Copyright (c) 2022 Siemens AG Oesterreich
+% SPDX-License-Identifier: MIT
+
+#program domain(new_id).
+
+
+% ooasp_arity(CONFIG, ASSOC, OPT, N, ARITY, STEP)
+% arity of node N for association as class C(OPT) is ARITY in step STEP
```
rtaupe:
```suggestion
% ooasp_arity(CONFIG, ASSOC, POS, N, ARITY, STEP)
% arity of association ASSOC for object N of class at position POS is ARITY in step STEP
```

asp_interactive_configuration/ooasp/encodings/ooasp_aux_config.lp
```diff
@@ -0,0 +1,37 @@
...
+
+ooasp_isa(CONFIG,SUPER,new_id) :- 
+    ooasp_configuration(V,CONFIG),
+    ooasp_isa(CONFIG,SUB,new_id),
+    ooasp_subclass(V,SUB,SUPER).
+
+% Auxiliary predicate to avoid duplicated rules
+% ooasp_associated_general(CONFIG,ASSOC,OPT,ID1,ID2)
+%    Object ID1 apearing in position OPT of association ASSOC is associated to object ID2
+%    ooasp_associated_general(c1,ass,2,23,10) object 23 is the class C2 of association ass with object 10
```
rtaupe:
```suggestion
% Auxiliary predicate to avoid duplicated rules
% ooasp_associated_general(CONFIG,ASSOC,POS,ID1,ID2)
%    Object ID1 apearing in position POS of association ASSOC is associated to object ID2
%    Example: ooasp_associated_general(c1,ass,2,23,10): : object 23 of the second class of association ass is associated with object 10
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -0,0 +1,148 @@
+% Copyright (c) 2022 Siemens AG Oesterreich
+% SPDX-License-Identifier: MIT
+
+#program domain(new_id).
+
+%------ Errors into constraints
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -0,0 +1,148 @@
...
+
+ooasp_cv(CONFIG,lowerbound,ID1,"Lowerbound for association {} not reached: {}",(ASSOC,CMIN,new_id)):-
+	ooasp_configuration(V,CONFIG),
+	ooasp_assoc_limit(V,ASSOC,min,OPT,C,CMIN),
+	ooasp_isa(CONFIG,C,ID1),
+	#count { ID2:ooasp_associated_general(CONFIG,ASSOC,OPT,ID1,ID2) } < CMIN,
+	active(new_id).
+
+
+%--- Associations are only allowed to something instanciated
```
rtaupe:
```suggestion
%--- Associations are only allowed to something instantiated
```

asp_interactive_configuration/ooasp/encodings/ooasp_guess.lp
```diff
@@ -0,0 +1,47 @@
...
+
+% Attributes
+1 { ooasp_attribute_value(CONFIG,N,new_id,VALUE):ooasp_attribute_enum(V,C,N,VALUE) } 1 :-
+	ooasp_isa(CONFIG,C,new_id),
+	ooasp_attribute(V,C,N,T),
+	ooasp_attribute_hasdomain(V,C,N),
+	ooasp_configuration(V,CONFIG),
+	guess.
+
+% TODO: this rule is only used if MAX-MIN < 30
```
rtaupe, line 39: This seems to be obsolete since line 45 is commented out.

asp_interactive_configuration/ooasp/config.py
```diff
@@ -0,0 +1,386 @@
...
+from typing import List
+from types import SimpleNamespace
+from clingo import Model, parse_term
+from clorm.clingo import Control
+from clorm import Symbol, Predicate, ConstantField, IntegerField, FactBase, RawField, refine_field, parse_fact_files, Raw, StringField, unify
+from clingraph.orm import Factbase
+from clingraph.graphviz import compute_graphs, render
+from .kb import OOASPKnowledgeBase
+from copy import deepcopy, copy
+import ooasp.utils as utils
```
rtaupe:
```suggestion
from typing import List
from types import SimpleNamespace
from clingo import Model, parse_term
from clorm.clingo import Control
from clorm import Symbol, Predicate, ConstantField, IntegerField, FactBase, RawField, refine_field, Raw, StringField
from clingraph.orm import Factbase
from clingraph.graphviz import compute_graphs, render
from .kb import OOASPKnowledgeBase
from copy import deepcopy
import ooasp.utils as utils
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+from textwrap import dedent
+from clorm.clingo import Control
+from clingo import Symbol, Number, Function
+from clorm import Predicate, unify
+from ooasp.config import OOASPConfiguration
+from ooasp.kb import OOASPKnowledgeBase
+from typing import List 
+from copy import copy, deepcopy
+import ooasp.utils as utils
+from typing import List
```
rtaupe:
```suggestion
import time
from clorm.clingo import Control
from clingo import Number, Function
from clorm import Predicate, unify
from ooasp.config import OOASPConfiguration
from ooasp.kb import OOASPKnowledgeBase
from typing import List 
from copy import deepcopy
import ooasp.utils as utils
from typing import List
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+    def __init__(self, config:OOASPConfiguration,
+                        action:str,
+                        domain_size:int):
+        """
+        Creates a state
+            Parameters:
+                kb: The Knowledge base
+                config: The partial or complete configuration
+                domain_size: The number of objects in the partial configuration
+                action: The action perfored to get the state
```
rtaupe:
```suggestion
                action: The action performed to get the state
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        return cls(OOASPConfiguration(config_name,kb),
+            action="start",
+            domain_size=0)
+
+    def copy(self, action:str, deep_config:bool = False):
+        """
+        Creates a state by copying the state and changing some properties
+            Parameters:
+                action: The action for the new state
+                deep_config: Wether the configuration should be copied in a deep way
```
rtaupe:
```suggestion
                deep_config: Whether the configuration should be copied in a deep way
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        self.ctl.ground(args)
+        end = time.time()
+        self._add_grounding_time(end -start)
+
+
+    def _ground_missing(self)->None:
+        """
+        Grounds all missing programs based on the current domain_size
+        """
+        # print("Ground missing")
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        """
+        for f in facts:
+            self.ctl.assign_external(Function("user",[f.symbol]), False)
+
+    def _add_fact(self, fact:Predicate)->None:
+        """
+        Adds a new fact to the control. The fact is added as part
+        of the program for the current domain size.
+            Parameters:
+                facts: The clorm predicate to be added to the control
```
rtaupe:
```suggestion
                fact: The clorm predicate to be added to the control
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+    # --------- Browsing
+
+    def _next_solution(self)->OOASPConfiguration:
+        """
+        Gets the next avaliable solution for the current configuration without increasing the domain size.
+        If there is an ongoing browsing process it will continue ussing
+        the given iterarator, otherwise tt will ground any missing steps and solve.
+        Sets the found configuration based on the next computed model.
+            Returns:
+                The found OOASPConfiguration or None if no solution is found
```
rtaupe:
```suggestion
        Gets the next available solution for the current configuration without increasing the domain size.
        If there is an ongoing browsing process it will continue using
        the given iterator, otherwise it will ground any missing steps and solve.
        Sets the found configuration based on the next computed model.
            Returns:
                The found OOASPConfiguration or None if no solution is found
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+
+        if not self.browsing:
+            self.ctl.assign_external(Function("guess"), True)
+            self.ctl.assign_external(Function("check"), True)
+            self.ctl.assign_external(Function("check_partial_cv"), True)
+
+            self._set_user_externals()
+            self.ctl.configuration.solve.enum_mode = 'auto'
+            start = time.time()
+            # print("\n---Calling handle")
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+            self.hdn = self.ctl.solve(yield_=True)
+            end = time.time()
+            self._add_solving_time(end -start)
+
+            self.solution_iterator = iter(self.hdn)
+        
+        start = time.time()
+        try:
+            # print("---Calling next")
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+            model = next(self.solution_iterator)
+            # print("Done")
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+            end = time.time()
+            self._add_solving_time(end -start)
+            found_config = OOASPConfiguration.from_model(self.state.config.name,
+                    self.kb, model)
+            return found_config
+        except StopIteration:
+            # print("Cancel")
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+    def _brave_config_as_options(self) -> dict:
+        """
+        Returns the brave configuration computed as a dictionary with 
+        the options per object
+            Returns:
+                The dictionary with ids of objects as keys and list of dicionaries as value 
+                representing the possible options
+        """
+        if self.brave_config is None:
+            raise RuntimeError("A brave configuration must be computed to get it as optionss")
```
rtaupe:
```suggestion
            raise RuntimeError("A brave configuration must be computed to get it as options")
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        Gets the next avaliable solution for the current configuration without increasing the domain size.
+        If there is an ongoing browsing process it will continue ussing
+        the given iterarator, otherwise tt will ground any missing steps and solve.
+        Sets the found configuration based on the next computed model.
+            Returns:
+                The found OOASPConfiguration or None if no solution is found
+        """
+        self._ground_missing()
+
+        if not self.browsing:
```
rtaupe, line 288: Maybe there's something about Python I don't know, but shouldn't the method `self.browsing` be called with parentheses? (Here and in other places)

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        self._ground_missing()
+        self.ctl.assign_external(Function("guess"), False)
+        self.ctl.assign_external(Function("check"), False)
+        self.ctl.assign_external(Function("check_partial_cv"), False)
+        self._set_user_externals()
+        self.ctl.configuration.solve.enum_mode = 'auto'
+        with  self.ctl.solve(yield_=True) as hdn:
+            for model in hdn:
+                # for s in model.symbols(atoms=True):
+                    # print(s)
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        self._new_state("Obtaining options")
+        self.found_config=found_config
+        return self._get_options()
+
+    def next_solution(self)->OOASPConfiguration:
+        """
+        Creates a new state.
+        Gets the next avaliable solution for the current configuration without increasing the domain size.
+        If there is an ongoing browsing process it will continue ussing
+        the given iterarator, otherwise tt will ground any missing steps and solve.
```
rtaupe:
```suggestion
        If there is an ongoing browsing process it will continue using
        the given iterator, otherwise it will ground any missing steps and solve.
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+    
+    def select_leaf_class(self,object_id:int,leaf_class:str)->None:
+        """
+        Creates a state with a new configuration.
+        Removes any current leaf selection for the object.
+        Selects the leaf class for the object.
+            Parameters:
+                object_id: Id of the object
+            Throws:
+                Error in case the leaf_class is not really a leaf
```
rtaupe:
```suggestion
                Error in case the leaf_class is not really a leaf class
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -0,0 +1,710 @@
...
+        """
+        Creates a state with a new configuration.
+        Trys to find a solution for the current domain size, if a solution is
+        found it is returned otherwise it will increase the domain by one
+        and look for solution until one is found.
+        Sets the found configuration based on the next computed model.
+        The found configuration can be selected as the new one using select_found_configuration
+
+        This will affect the Number of object in the current configuration
+        even if the found condiguration is never selected.
```
rtaupe:
```suggestion
        even if the found configuration is never selected.
```

asp_interactive_configuration/ooasp/tests/test_config.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2022 Siemens AG Oesterreich
+# SPDX-License-Identifier: MIT
+
+from clorm.clingo import Control
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/tests/test_correctness.py
```diff
@@ -0,0 +1,130 @@
+# Copyright (c) 2022 Siemens AG Oesterreich
+# SPDX-License-Identifier: MIT
+
+from ooasp.interactive import InteractiveConfigurator
+from ooasp.kb import OOASPKnowledgeBase
+from time import time
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/tests/test_interactive.py
```diff
@@ -0,0 +1,214 @@
+# Copyright (c) 2022 Siemens AG Oesterreich
+# SPDX-License-Identifier: MIT
+
+import pytest
+from clorm.clingo import Control
+from clingo import parse_term
+from ooasp.kb import OOASPKnowledgeBase
+from ooasp.config import OOASPConfiguration
+from ooasp.interactive import InteractiveConfigurator
```
rtaupe:
```suggestion
import pytest
from clingo import parse_term
from ooasp.kb import OOASPKnowledgeBase
from ooasp.interactive import InteractiveConfigurator
```

asp_interactive_configuration/ooasp/tests/test_performance.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2022 Siemens AG Oesterreich
+# SPDX-License-Identifier: MIT
+
+from ooasp.interactive import InteractiveConfigurator
+from ooasp.kb import OOASPKnowledgeBase
+from time import time
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/tests/test_performance.py
```diff
@@ -0,0 +1,36 @@
...
+
+    racks_kb = OOASPKnowledgeBase.from_file("racks_v1","./examples/racks/kb.lp")
+    iconf = InteractiveConfigurator(racks_kb,"i1",["./examples/racks/constraints.lp"])    
+    # current limit
+    nr_of_elements = 18
+    for i in range(nr_of_elements):
+        iconf.new_leaf("elementA")
+    iconf.extend_domain(nr_of_elements + 5)
+    config = iconf.next_solution()
+    # config.save_png()
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/ui.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __init__(self, iconf):
+        """
+        Creates the UI
+        """
+        self.iconf = iconf
+        self.config_image = widgets.Image(format='png')
+        self.found_config = widgets.Image(format='png')
+
+        self.history = VBox(layout=Layout(height='300ox', width='auto',overflow_y='auto'))
```
rtaupe, line 28: Is `ox` a typo, should it be `px`?

asp_interactive_configuration/ooasp/ui.py
```diff
@@ -0,0 +1,223 @@
...
+        grid[1, 2:4] = self.edit
+        grid[1, 4:] = self.extend
+        grid[2, 0] = self.browse
+        grid[2, 1:] = self.found_config
+        self.grid = grid
+
+
+    def update(self):
+        """
+        Uodates the UI based on the current interactive state
```
rtaupe:
```suggestion
        Updates the UI based on the current interactive state
```

asp_interactive_configuration/ooasp/utils.py
```diff
@@ -0,0 +1,69 @@
...
+    return COLORS['BLUE'] + "="*10 + s + "="*10 + COLORS['NORMAL'] + "\n"
+
+def subtitle(s,color='CYAN'):
+    return COLORS[color] + "-"*10 + s + "-"*10 + COLORS['NORMAL'] + "\n"
+
+def pretty_dic(d):
+    s = (json.dumps(d, indent=4))
+    # s = ""
+    # for k,v in d.items():
+    #     s += COLORS['GREY'] + k + " : " + str(v) + COLORS['NORMAL'] + "\n"
```
rtaupe:
```suggestion
```

asp_interactive_configuration/ooasp/ui.py
```diff
@@ -0,0 +1,223 @@
+# Copyright (c) 2022 Siemens AG Oesterreich
+# SPDX-License-Identifier: MIT
+
+import ipywidgets as widgets
+from ipywidgets import Button, HBox, VBox, Label, Layout,GridspecLayout, HTML
```
rtaupe:
```suggestion
from ipywidgets import Button, VBox, Label, Layout,GridspecLayout, HTML
```

asp_interactive_configuration/pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[tool.poetry]
+name = "asp_interactive_configuration"
+version = "0.1.0"
+description = "project to develop an interactive asp-based configuration approach"
+authors = ["gschennersie"]
```
rtaupe, line 5: According to https://python-poetry.org/docs/pyproject/, authors must be in the form `name <email>`.

asp_interactive_configuration/usage/Configuration.ipynb
```diff
@@ -0,0 +1,252 @@
...
+     "execution_count": 9,
+     "metadata": {},
+     "output_type": "execute_result"
+    },
+    {
+     "ename": "",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31mThe Kernel crashed while executing code in the the current cell or a previous cell. Please review the code in the cell(s) to identify a possible cause of the failure. Click <a href='https://aka.ms/vscodeJupyterKernelCrash'>here</a> for more info. View Jupyter <a href='command:jupyter.viewOutput'>log</a> for further details."
```
rtaupe, line 199: I suggest to re-run the notebook to see if the error disappears

asp_interactive_configuration/usage/KB.ipynb
```diff
@@ -0,0 +1,380 @@
...
+     "execution_count": 13,
+     "metadata": {},
+     "output_type": "execute_result"
+    },
+    {
+     "ename": "",
+     "evalue": "",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31mThe Kernel crashed while executing code in the the current cell or a previous cell. Please review the code in the cell(s) to identify a possible cause of the failure. Click <a href='https://aka.ms/vscodeJupyterKernelCrash'>here</a> for more info. View Jupyter <a href='command:jupyter.viewOutput'>log</a> for further details."
```
rtaupe, line 328: I suggest to re-run the notebook to see if the error disappears

rtaupe (changes requested):
@gschennersie, thank you very much for committing this contribution!
Could you please look through my comments before we merge?

rtaupe (changes requested):
@gschennersie, thank you very much for committing this contribution!
Could you please look through my comments before we merge?

rtaupe:
Issues created while reviewing this PR that need not be fixed before the merge:
#6 
#7 
#8

asp_interactive_configuration/pyproject.toml
gschennersie: I don't want to provide my email.

asp_interactive_configuration/examples/metro/constraints.lp
gschennersie: The metro example is still WIP.

asp_interactive_configuration/examples/metro/constraints.lp
gschennersie: The metro example is work in progress and will be fixed by #9

asp_interactive_configuration/examples/racks/constraints.lp
gschennersie: If  using active(new_id) is enough for checking the constraint or if there need to be multiple cv rules.

asp_interactive_configuration/examples/racks/constraints.lp
gschennersie: These and similar bugs will be fixed in #10

asp_interactive_configuration/ooasp/config.py
gschennersie: This and similar will be checked in #11

asp_interactive_configuration/ooasp/ui.py
gschennersie: #11

gschennersie (commented):
Review finished. Created open issues. Ready to merge.

asp_interactive_configuration/usage/Configuration.ipynb
gschennersie: #11

asp_interactive_configuration/usage/KB.ipynb
gschennersie: #11

gschennersie (commented):
review completed.

rtaupe (approved):

Issue #7, opened 2023-01-16
Revise terminology in Interactive Configuration API


rtaupe:
The Interactive Configuration API introduced by #5 contains some terminology we want to revise:

- [ ] the term "leaf" or "leaf class" is sometimes not used correctly
- [ ] complete/partial constraints: this naming is not very clear and could be revised if we have a good idea

PR #16, opened 2023-05-10
Sh/updates

susuhahnml:


gschennersie (dismissed):
For me everything looks ok and the test run without errors.

PS: Some minor point most probably unrelated to this branch: During checking your branch i also checked the Jupyter Notebooks and in usage\Configuration.ipynb and usage\Interactive.ipynb there is an error when i run the last cells at least on my computer. If you have any ideas to automatically test Notebooks let me know. We have this problem in other projects. And when visualizing configurations the classnames are in upper case (e.g. Frame) but for the kb they are in lowercase.

asp_interactive_configuration/DOC.md
```diff
@@ -106,20 +106,20 @@ Constraints are defined using the `ooasp_cv/5` predicate:
 
 #### Complete constraints
```
rtaupe: If I understand correctly, constraints previously called "complete" are now called "permanent", and constraints previously called "partial" are now called "potential". If that is true, all occurrences of the old terms should be updated (e.g., here).

Furthermore, I think the new terminology is an improvement (#7), but maybe we still find a better word than "potential", e.g., "temporary" (as an antonym to "permanent"). But for now we can keep "potential".

asp_interactive_configuration/ooasp/encodings_simple/ooasp.lp
```diff
@@ -7,8 +7,8 @@
 ooasp_configuration(kb_name,config_name).
 #include "ooasp_aux_kb.lp". % Auxiliary predicates for the KB
 #external guess. % The guessing part is active
-#external check_partial_cv. % Checks for partial constraints apply
-#external check. % Checks for full constraints apply
+#external check_potential_cv. % Checks for potential constraints apply
+#external check_permanent_cv. % Checks for full constraints apply
```
rtaupe:
```suggestion
#external check_permanent_cv. % Checks for permanent constraints apply
```

asp_interactive_configuration/ooasp/tests/test_performance.py
```diff
@@ -14,7 +14,13 @@ def test_solve_multiple_elements():
     for i in range(nr_of_elements):
         iconf.new_leaf("elementA")
     iconf.extend_domain(nr_of_elements + 5)
+    # print(iconf._statistics)
     config = iconf.next_solution()
+    # print(iconf._statistics)
+    iconf.check()
+    # print(iconf._statistics)
+
+
```
rtaupe: please remove temporarily needed code before merging

rtaupe (changes requested):
Thank you, Susana!
Please let's just address a minor issue regarding new terminology before merging.

asp_interactive_configuration/ooasp/encodings_simple/ooasp.lp
susuhahnml: I added it in my commit

susuhahnml:
All addressed in the last commit!

rtaupe:
> All addressed in the last commit!

The place I pointed out where the old terminology is still used was just an example, there are also others. Do you want to fix them or shall I?

susuhahnml:
I just updated the ones I could find, if you have any more in the radar, please add them :)

rtaupe (approved):

rtaupe:
Thank you!

PR #22, opened 2023-08-04
Refactored _brave_config_as_options and added Docstrings

ThmsMoritz:
Extracted new methods from _brave_config_as_options and wrote the respective Docstrings, as well as added some previously missing ones in ui.py and interactive.py.

### DIFF
--- a/asp_interactive_configuration/ooasp/interactive.py
+++ b/asp_interactive_configuration/ooasp/interactive.py
@@ -383,7 +383,7 @@ class InteractiveConfigurator:
         self.state.config=config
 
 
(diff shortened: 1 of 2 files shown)


susuhahnml:
looking good :) Dont forget to make sure the tests run:
`pytest ooasp/tests`

rtaupe (changes requested):
Thank you!
Please just rename the methods to use `dict` instead of `dic`, since the former is a more common abbreviation for "dictionary".

gschennersie (approved):

PR #24, opened 2023-08-07
Subobj

susuhahnml:


gschennersie (approved):

asp_interactive_configuration/benchmarks/benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+        results.append(BM(n_runs,e,fun,ne=e))
+
+    save_results(results,name)
+
+# run(2,extend_solve,elements=[13,14,15,16,17,18],name=f"{opt}-os/extend_solve")
+# run(2,incremental,elements=[8,9,10,11],name=f"{opt}/incremental-noex")
+run(2,incremental,elements=[8,9,10,11],name=f"{opt}-os/incremental")
+# run(2,options,elements=[18,20,22,24],name=f"{opt}-os/options")
+# run(2,options,elements=[20,22,24,25,26],name=f"{opt}/options")
+# run(2,options_object,elements=[18,20,22,24],name=f"{opt}/options_object")
```
rtaupe: Please don't include commented-out code, because for readers of the code it will be difficult to understand what it means and it will also be hard to maintain.

asp_interactive_configuration/examples/metro/constraints.lp
rtaupe, line 1: Why was the metro example deleted?

asp_interactive_configuration/ooasp/config.py
```diff
@@ -380,22 +439,18 @@ def save_png(self,directory:str="./out"):
...
+        path = settings.encodings_path.joinpath("ooasp_aux_kb.lp")
+        ctl.load(str(path))
 
         ctl.add("base",[],self.fb.asp_str())
         ctl.add("base",[],self.kb.fb.asp_str())
         ctl.ground([("base", [])],ClingraphContext())
         ctl.solve(on_model=lambda m: fbs.append(Factbase.from_model(m,default_graph="config")))
         graphs = compute_graphs(fbs[0])
+        # g = unflatten(str(graphs['config']),chain=2)
+        # graphs['config']=Source(g)
```
rtaupe: Please don't include commented-out code, because for readers of the code it will be difficult to understand what it means and it will also be hard to maintain.

asp_interactive_configuration/ooasp/encodings_simple/ooasp.lp
rtaupe, line 1: For the commit history, it would have been nicer to simply rename `encodings_simply` to `encodings` instead of deleting the former and modifying the latter. But I guess now it would be too much effort to revert this decision.

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -12,8 +12,11 @@
 from copy import deepcopy
 import ooasp.utils as utils
 from typing import List
-import ooasp.encodings
-import ooasp.encodings_simple
+import ooasp.settings as settings
+
+def log(x):
+    # print(x)
+    return
```
rtaupe: Please add a docstring, e.g., "Used for logging if print(x) is not commented out"

rtaupe (changes requested):
Thank you for these contributions!

Please don't include commented-out code, because for readers of the code it will be difficult to understand what it means and it will also be hard to maintain. This concerns several files!

Please also address my other comments briefly.

asp_interactive_configuration/benchmarks/benchmark.py
```diff
@@ -0,0 +1,153 @@
...
+    n_runs = 3
+    results = []
+    for e in elements:
+        results.append(BM(n_runs,e,fun,ne=e))
+
+    save_results(results,name)
+
+# run(2,extend_solve,elements=[13,14,15,16,17,18],name=f"{opt}-os/extend_solve")
+# run(2,incremental,elements=[8,9,10,11],name=f"{opt}/incremental-noex")
+run(2,incremental,elements=[8,9,10,11],name=f"{opt}-os/incremental")
```
rtaupe: When I run this file, I get errors (`opt` is not defined, `settings` is not defined, ...).
Is there an import missing? Or how is this script supposed to be called?

asp_interactive_configuration/examples/metro/constraints.lp
susuhahnml: Probably intended to move it since it was outdated. I just added it again

asp_interactive_configuration/benchmarks/benchmark.py
susuhahnml: I added some docstrings to these files and remove the commented lines. The lines in this case where the benchmarks that were ran, no the files only hold the functions.

asp_interactive_configuration/ooasp/encodings_simple/ooasp.lp
susuhahnml: Yes, right now, as we discussed before I removed all the different versions that where hard to mantain. It is not clear how to do this with multiple encodings. I will try to keep this modular for benchmarking different options latter on.

asp_interactive_configuration/benchmarks/benchmark.py
susuhahnml: I removed this opt together with the different versions. Now this file is up to date but only with the utils for benchmarking. We latter on need to decide what to benchmark.

rtaupe (approved):

PR #33, opened 2023-10-18
Cautious

susuhahnml:


asp_interactive_configuration/DOC.md
```diff
@@ -26,7 +33,8 @@ The following tasks correspond to the only points of the interactive process whe
 
 1. *Complete:* Getting a complete configuration extending `C`.
 2. *Check:* Checking `C` for errors.
-3. *Options:* Obtaining all possible classes for the objects, values for the attributes, and associations to complete `C`.
+3. *Options:* Obtaining all possible classes, values, and associations to complete `C`.
```
rtaupe: For 3., I find the old version of the text better understandable. If it is still correct, I suggest reverting back to it.

asp_interactive_configuration/DOC.md
```diff
@@ -130,17 +138,73 @@ The user input corresponds to the current partial configuration `C` that is bein
...
+- **`user(ooasp_associated(ASSOC,ID1,ID2))`**
+  The user associated `ID1` and `ID2` via `ASSOC`
 
-- **`user(ooasp_attribute_value(CONFIG,N,ID,VALUE))`**
+- **`user(ooasp_attribute_value(N,ID,VALUE))`**
   The user selected value `VALUE` for attribute `N` of object `ID`
 
 The truth value of these externals is defined by the [Interactive Configurator](#interactive-configurator) based on the current (partial) configuration that is being constructed using an object of class [OOASPConfiguration](#configuration).
 
+The user externals are generated restricted by the possible classes that an object can have. This is defined by the argument `cls` passed in the grounding, which corresponds to the domain defined by `ooasp_domain(_,cls).`. Therefore, if the new object is grounded for a class `cls`, only valid attributes and associations for `cls` and its subclasses will be considered. This is summarized in the auxiliary predicate `counts_as(O,C)`, which states that object `O` can count as class `C`.
```
rtaupe: `ooasp_domain` is mentioned nowhere else in DOC.md, which makes this explanation a bit hard to understand. Furthermore, looking at usages of this predicate in the encoding, the class seems to be the first argument, not the second. Is `ooasp_domain(_,cls)` correct then?

asp_interactive_configuration/DOC.md
```diff
@@ -130,17 +138,73 @@ The user input corresponds to the current partial configuration `C` that is bein
...
+
+
+## Numerical attributes
+
+Numerical attributes are treated using the system `fclingo`. This systems adds a founded condition while using `clorm` in the background. 
+Attributes whose type is `int`, such as `nr_passengers`, `standing_room`, and `nr_seats` in the Metro example, are treated as variables with this system as explained below. Note that numerical values can also be treated as an enumeration using the type `enumint` instead. This option will consider all values in the defined range for predicate `ooasp_attr_value`, such as `frame_position` in the racks example.
+
+### Usage
+
+When an attribute `ATTR` is of type `int`, their value is defined using an `fclingo` variable `ooasp_attr_fvalue(ATTR,O)`.  The value of such variable will be associated to the value of attribute `ATTR` for object `O`.
```
rtaupe:
```suggestion
When an attribute `ATTR` is of type `int`, their value is defined using an `fclingo` variable `ooasp_attr_fvalue(ATTR,O)`.  The value of such a variable will be associated to the value of attribute `ATTR` for object `O`.
```

asp_interactive_configuration/DOC.md
```diff
@@ -130,17 +138,73 @@ The user input corresponds to the current partial configuration `C` that is bein
...
+These values can be compared using the theory symbol `&fsum`. For instance, in the following rule we use `&fsum{ooasp_attr_fvalue(A,new_object)}<MIN` to state that the value of `A` is smaller than `MIN` leading to a constraint violation. Notice that the variable `MIN` must be defined in a positive literal of the body.
```
rtaupe:
```suggestion
These values can be compared using the theory symbol `&fsum`. For instance, in the following rule we use `&fsum{ooasp_attr_fvalue(A,new_object)}<MIN` to state that the value of `A` being smaller than `MIN` will lead to a constraint violation. Notice that the variable `MIN` must be defined in a positive literal of the body.
```

asp_interactive_configuration/DOC.md
```diff
@@ -130,17 +138,73 @@ The user input corresponds to the current partial configuration `C` that is bein
...
+To check if a value is defined one can do so by checking if it is equal to itself. In the case of the next rule, if no value is defined, then the constraint violation is inferred.
+
+```
+ooasp_cv(no_value,new_object,"Missing value for {}",(A,)) :-
+	ooasp_attr(C,A,int),
+	ooasp_isa(C,new_object),
+	not &fsum{ooasp_attr_fvalue(A,new_object)}=ooasp_attr_fvalue(A,new_object).
+```
+
+This type of attributes must be treated differently to the rest, therefore they require rules like to ones above to handle each case. Below we show the difference of writing a constraint violation using `ooasp_attr_value/3` and `ooasp_attr_fvalue/2`. Notice that in the second case, we directly compare the values by summing them up.
```
rtaupe:
```suggestion
This type of attributes must be treated differently to the rest, therefore they require rules like the ones above to handle each case. Below we show the difference of writing a constraint violation using `ooasp_attr_value/3` and `ooasp_attr_fvalue/2`. Notice that in the second case, we directly compare the values by summing them up.
```

asp_interactive_configuration/DOC.md
```diff
@@ -130,17 +138,73 @@ The user input corresponds to the current partial configuration `C` that is bein
...
+          ooasp_attr_fvalue(standing_room,new_object)}!=
+        ooasp_attr_fvalue(nr_passengers,new_object).
+```
+
+### Interactivity
+
+The assigned values for these attributes are handled internally by the `fclingo `system and are only accessed after the models are found. Therefore, they are not considered in any brave or cautious consequences. This means that we do not know in advance the possible values an attribute can take for a partial configuration. Interactively, this can be solved in different ways, such as using a slider instead of a dropdown or some other numerical input. For convenience, we keep the usage in the jupyter notebooks with dropdowns and add all possible values for the attribute, even though they might create an invalid configuration. Invalid configurations can still be checked and fixed in the UI. 
+
+In this version, externals are used for every possible value to allow user input. However, this is not efficient. Another version is proposed in branch `efficient-fclingo` where externals are added on demand. However. This option does not allow for the task `check`, since it requires the choices to be active.
```
rtaupe:
```suggestion
In this version, externals are used for every possible value to allow user input. However, this is not efficient. Another version is proposed in branch `efficient-fclingo` where externals are added on demand. However, this option does not allow for the task `check`, since it requires the choices to be active. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -162,30 +225,28 @@ Take for instance the following rule:
 
 ```prolog
 #program domain(new_object).
```
rtaupe: Shouldn't this be `#program domain(new_object, cls).`?

asp_interactive_configuration/DOC.md
```diff
@@ -196,47 +257,45 @@ There are some rules that depend on atoms that are computed in an accumulative w
 Consider the following rule:
 
 ```prolog
-arity(new_object,ASSOC,ARITY):- ARITY = #count{ID2:ooasp_assoc(V,ASSOC,new_object,_,_,ID2,_,_)}.
+arity(new_object,ASSOC,ARITY):- ARITY = #count{ID2:ooasp_assoc(ASSOC,new_object,_,_,ID2,_,_)}.
```
rtaupe:
```suggestion
arity(new_object,ASSOC,ARITY) :- ARITY = #count{ID2:ooasp_assoc(ASSOC,new_object,_,_,ID2,_,_)}.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+- **`ooasp_associated_general(ASSOC,POS,ID1,ID2)`**
   Object `ID1` appearing in position `POS` of association `ASSOC` is associated to object `ID2`
   *Example: `ooasp_associated_general(c1,ass,2,23,10)`: object `23` of the second class of association `ass` is associated with object `10`*
 
-### Incremental solving
+--- 
+
+## Incremental solving
+
+This multi-shot encoding can directly be used for incremental solving. With this idea we ground for the current objects and try to find a configuration using Task 1 *Complete*, if there is no complete configuration (returns UNSAT) with this number of objects, we ground for one more object and repeat the process until we have a satisfiable answer.
```
rtaupe:
```suggestion
This multi-shot encoding can directly be used for incremental solving. With this idea we ground for the current objects and try to find a configuration using Task 1 *Complete*, and if there is no complete configuration (returns UNSAT) with this number of objects, we ground for one more object and repeat the process until we have a satisfiable answer.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+This incremental process find the smallest configuration. However, it has the drawback that right before getting the satisfiable answer the solving time increases since it has to look over the whole search space, which impacts in performance.
```
rtaupe:
```suggestion
This incremental process finds the smallest configuration. However, it has the drawback that right before getting the satisfiable answer the solving time increases since the solver has to look over the whole search space, which impacts its performance.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+Additionally, if there are cases where no configuration can be found, the incremental solving will continue adding objects util the set limit is reached. For instance, in the metro example, if one Wagon is added with number of passengers 0, and a seat object is added, then this partial configuration can not reach a compleat configuration, since it would require two wagons, which is not possible. However, the incremental solving would not know this and continue looking for the answer.
```
rtaupe:
```suggestion
Additionally, if there are cases where no configuration can be found, the incremental solving will continue adding objects util the set limit is reached. For instance, in the metro example, if one Wagon is added with number of passengers 0, and a seat object is added, then this partial configuration can not reach a complete configuration, since it would require two wagons, which is not possible. However, the incremental solving would not know this and continue looking for the answer.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+
+###  Create required
+
+Creates all the objects required by the current objects. To do so, we use the cautious consequences of the program for getting the constraint violations that are common to all models, and proceed to solve one of them at a time. For this process we need optimization statements, to get rid of models that include unnecessary  constraint. This optimization is only activate for getting cautious consequences.
```
rtaupe: What are models that "include unnecessary constraint[s]"?

rtaupe:
```suggestion
Creates all the objects required by the current objects. To do so, we use the cautious consequences of the program for getting the constraint violations that are common to all models, and proceed to solve one of them at a time. For this process we need optimization statements, to get rid of models that include unnecessary constraint. This optimization is only active for getting cautious consequences. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+Additionally, if there are cases where no configuration can be found, the incremental solving will continue adding objects util the set limit is reached. For instance, in the metro example, if one Wagon is added with number of passengers 0, and a seat object is added, then this partial configuration can not reach a compleat configuration, since it would require two wagons, which is not possible. However, the incremental solving would not know this and continue looking for the answer.
+
+
+###  Create required
+
+Creates all the objects required by the current objects. To do so, we use the cautious consequences of the program for getting the constraint violations that are common to all models, and proceed to solve one of them at a time. For this process we need optimization statements, to get rid of models that include unnecessary  constraint. This optimization is only activate for getting cautious consequences. 
+
+#### Optimize number of lower bond violations
```
rtaupe:
```suggestion
#### Optimize number of lower bound violations
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+The main optimization with priority `3`, penalizes lower bound violations based on how many objects have not yet been associated `CMIN-N`. This week constraint will prioritize models that associate as many objects as possible. For instance, take a partial configuration with one rack and 3 frames. Since the external `ooasp_potential_cv` is false we would get a model for associating 0,1,2 and 3 frames. We this week constraint we will only get one where 3 frames are associated and the lower bound is only violated with one missing frame for the `rack_frames` association.
```
rtaupe:
```suggestion
The main optimization with priority `3` penalizes lower bound violations based on how many objects have not yet been associated `CMIN-N`. This weak constraint will prioritize models that associate as many objects as possible. For instance, take a partial configuration with one rack and 3 frames. Since the external `ooasp_potential_cv` is false we would get a model for associating 0,1,2 and 3 frames. With this weak constraint we will only get one where 3 frames are associated and the lower bound is only violated with one missing frame for the `rack_frames` association.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+#### Optimize associations to smallest objects
+
+This optimization can be viewed like a symmetry constraint. The idea is to try to associate the smallest objects first by penalizing based on the ids of the objects associated. This is done in two constraints to consider both positions for the association. With this in place, we would make sure that if we have one rack and 5 frames, always the frame with the largest id would remain not associated, those making the same constraint violation appear in all stable models and consequently in the cautious consequences. Without this optimization, we would have a model for each frame being left out and no constraint violation would appear in the intersection. 
+
+
+#### Using the cautions consequences to create required objects
+
+1. Gets the cautious consequences for the program where potential cv can be violated and the optimization is activated.
+2. Takes the first lower bound constraint violation. This predicate already has the information of how many objects where associated and what the minimum was. Based on this information it crates and associates the required objects of the specific class
```
rtaupe:
```suggestion
2. Takes the first lower bound constraint violation. This predicate already has the information of how many objects were associated and what the minimum was. Based on this information it creates and associates the required objects of the specific class.
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+
+## Advanced features to simplify writing domain specific constraint violations
+
+### Association specialization
+
+Association specializations will specialize the lower and upper bound of an association for subclasses.
+
+For example, the association `rack_frames` between class `rack` and `frame` is defined by the atom `ooasp_assoc(rack_frames,rack,1,1,frame,4,8)`.
+The racks example limits the number of racks associated to the racks subclass `rackSingle` to be exactly `4`. This is achieved by adding an association specialization with name `rack_framesS` which specializes `rack_frames`, by the atom `ooasp_assoc_specialization(rack_framesS,rack_frames)`. This means that all `rack_frameS` associations are also `rack_frames` associations. Then, the upper and lowe bound can be redefined when the association is defined in `ooasp_assoc(rack_framesS,rackSingle,0,1,frame,4,4)`.
```
rtaupe:
```suggestion
The racks example limits the number of racks associated to the racks subclass `rackSingle` to be exactly `4`. This is achieved by adding an association specialization with name `rack_framesS` which specializes `rack_frames`, by the atom `ooasp_assoc_specialization(rack_framesS,rack_frames)`. This means that all `rack_framesS` associations are also `rack_frames` associations. Then, the upper and lower bound can be redefined when the association is defined in `ooasp_assoc(rack_framesS,rackSingle,0,1,frame,4,4)`. 
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+
+Specialized constraint violations are those used in different instances but not encoded in the knowledge base. Some of them are the following:
+
+- **(unique)** all attribute values of objects in an association are different 
+- **(same)** all attribute values of objects in an association are equal (exceptions)
+- **(table)** which value combinations are allowed
+- **(specialization)** attribute value determines the type of objects
+
+These constraints can be defined in the knowledge base using the predicate `ooasp_special_cv/2`, where the first argument is the name of the constraint and the second one has any additional information needed. To implement a specialized constraint violation, the file [ooasp/encodings/ooa_checksp.lp](ooasp/encodings_check/ooasp.lp) must be extended with the needed constraints.
+This process was done for the  **unique** constraint violation. It is used in the knowladge base of the racks example, to make sure that the frame positions of all frames associated to a rack are different. In this case the second argument is a tuple `(ASSOCIATION, CLASS, ATTRIBUTE)`
```
rtaupe:
```suggestion
This process was done for the  **unique** constraint violation. It is used in the knowledge base of the racks example, to make sure that the frame positions of all frames associated to a rack are different. In this case the second argument is a tuple `(ASSOCIATION, CLASS, ATTRIBUTE)`
```

asp_interactive_configuration/DOC.md
```diff
@@ -312,28 +371,111 @@ ooasp_potential_cv(wrongass_lower).
...
+### Specialized constraint violations
+
+Specialized constraint violations are those used in different instances but not encoded in the knowledge base. Some of them are the following:
+
+- **(unique)** all attribute values of objects in an association are different 
+- **(same)** all attribute values of objects in an association are equal (exceptions)
+- **(table)** which value combinations are allowed
+- **(specialization)** attribute value determines the type of objects
+
+These constraints can be defined in the knowledge base using the predicate `ooasp_special_cv/2`, where the first argument is the name of the constraint and the second one has any additional information needed. To implement a specialized constraint violation, the file [ooasp/encodings/ooa_checksp.lp](ooasp/encodings_check/ooasp.lp) must be extended with the needed constraints.
```
rtaupe:
```suggestion
### Special constraint violations

Special constraint violations are those used in different instances but not encoded in the knowledge base. Some of them are the following:

- **(unique)** all attribute values of objects in an association are different 
- **(same)** all attribute values of objects in an association are equal (exceptions)
- **(table)** which value combinations are allowed
- **(specialization)** attribute value determines the type of objects

These constraints can be defined in the knowledge base using the predicate `ooasp_special_cv/2`, where the first argument is the name of the constraint and the second one has any additional information needed. To implement a special constraint violation, the file [ooasp/encodings/ooa_checksp.lp](ooasp/encodings_check/ooasp.lp) must be extended with the needed constraints.
```
I propose to rename "specialized constraint violations" to "special constraint violations" to avoid confusion since, especially having read the section directly preceding this section, the reader could wrongly assume a conceptual connection to association specialization.

asp_interactive_configuration/poetry.lock
```diff
@@ -294,89 +294,183 @@ files = [
...
+
+[package.source]
+type = "legacy"
+url = "https://test.pypi.org/simple"
+reference = "test"
+
 [[package]]
 name = "clingo"
-version = "5.6.2"
+version = "5.7.0.post9"
```
rtaupe: `poetry install` gives me:

> RuntimeError
> Unable to find installation candidates for clingo (5.7.0.post9)

asp_interactive_configuration/ooasp/config.py
```diff
@@ -208,15 +219,23 @@ def leafs(self)->List[Predicate]:
...
     
     @property
     def smart_objects(self)->List[Predicate]:
         """
-        Smart objects
+        A list of objects of type self.UNIFIERS.Object such that only one instance for each object appears.
+        Internally uses the small_objects method and maps them into the Objects class
+
+        Returns:
+            A list of unique objects of type self.UNIFIERS.Object with their smallest class
```
rtaupe:
```suggestion
            A list of unique objects of type self.UNIFIERS.Object with their most specific instantiated class
```

asp_interactive_configuration/ooasp/config.py
```diff
@@ -254,17 +273,34 @@ def attribute_values(self)->List[Predicate]:
...
+        
+        return user_fb
 
-    def associated_by(self, obj, assoc_name)->List[Predicate]:
+    def associated_by(self, obj:int, assoc_name:str)->List[Predicate]:
         """
         The list of associations from object obj via association assoc_name
+
+        Parameters:
+            obj (int): Object where the association starts, can be right ot left
```
rtaupe: Is it correct that the integers 1 and 2 refer to the left and right end, respectively? If yes, please add this information here.

asp_interactive_configuration/ooasp/encodings/ooasp_aux_config.lp
```diff
@@ -71,4 +69,13 @@ ooasp_associated(ASSOC_S,ID1,new_object) :-
...
-    ooasp_isa(C2,new_object).
\ No newline at end of file
+    ooasp_isa(C2,new_object).
+
+
+% Int attributes
+
+ooasp_attr_int(new_object, A, MIN, MAX):-
+	ooasp_isa(C,new_object),
+    ooasp_attr_minInclusive(C,A,MIN),
```
rtaupe:
```suggestion
	ooasp_attr_minInclusive(C,A,MIN),
```

asp_interactive_configuration/ooasp/encodings/ooasp_aux_kb.lp
```diff
@@ -23,6 +25,11 @@ ooasp_attr_hasdomain(C,N) :-
     ooasp_attr_enum(C,N,D).
 
 ooasp_attr_enum(C,N,MIN..MAX):-
```
rtaupe:
```suggestion
ooasp_attr_enum(C,N,MIN..MAX) :-
```

asp_interactive_configuration/ooasp/encodings/ooasp_aux_kb.lp
```diff
@@ -23,6 +25,11 @@ ooasp_attr_hasdomain(C,N) :-
...
+    ooasp_attr(C,N,enumint),
+    ooasp_attr_minInclusive(C,N,MIN),
+    ooasp_attr_maxInclusive(C,N,MAX).
+
+ooasp_attr_fdom(C,N,MIN..MAX):-
```
rtaupe:
```suggestion
ooasp_attr_fdom(C,N,MIN..MAX) :-
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -95,13 +97,33 @@ ooasp_cv(wrongtypeinassoc,ID,"Associated by {} but is not of class {}",(ASSOC,C2
 % Values
 %=======================================
 
-%--- Unique attributes
+
+%--- Missing atribute value
```
rtaupe:
```suggestion
%--- Missing attribute value
```

asp_interactive_configuration/ooasp/encodings/ooasp_check.lp
```diff
@@ -95,13 +97,33 @@ ooasp_cv(wrongtypeinassoc,ID,"Associated by {} but is not of class {}",(ASSOC,C2
...
+ooasp_cv(no_value,new_object,"Missing value for {}",(A,)) :-
+	 ooasp_attr(C,A,enumint),
+	 ooasp_attr_enum(C,A,_),
+	 ooasp_isa(C,new_object),
+	 not ooasp_attr_value(A,new_object,_).
+
+%--- Missing atribute value (Integer)
```
rtaupe:
```suggestion
%--- Missing attribute value (Integer)
```

asp_interactive_configuration/ooasp/encodings/viz_config.lp
```diff
@@ -103,7 +103,7 @@ def cv_str(name,message,args):
     formatted_message = f"{message.string.format(*args_symbols)}"
     s_wrap_list = textwrap.wrap(formatted_message, 40)
     message_wraped = "<br ALIGN='left'/>   ".join(s_wrap_list)
-    s = f"<FONT COLOR='red'>(" + str(name) + ") " + "<I> "+ message_wraped  + "</I></FONT><br ALIGN='left'/> "
+    s = f"<FONT COLOR='red'><I> "+ message_wraped  + "</I> (" + str(name) + ") </FONT><br ALIGN='left'/> "
```
rtaupe:
```suggestion
    s = f"<FONT COLOR='red'><I> "+ message_wrapped  + "</I> (" + str(name) + ") </FONT><br ALIGN='left'/> "
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -114,13 +133,26 @@ def __init__(self, kb:OOASPKnowledgeBase, config_name:str, additional_files:List
         self.hdn = None
 
 
-    def _add_grounding_time(self, time):
+    def _add_grounding_time(self, time:int):
+        """
+        Adds time to grounding for benchamrks
```
rtaupe:
```suggestion
        Adds time to grounding for benchmarks
```

asp_interactive_configuration/ooasp/interactive.py
```diff
@@ -526,21 +608,36 @@ def _check(self)->bool:
...
     def _get_cautious(self)->OOASPConfiguration:
+        """
+        Gets the cautious consequences, also incudes an optimization statement 
+        to only consider the models where lowerbound violations are minimized
+
+        Raises:
+            RuntimeError: If the configuration is conflicting and has no models
+
+        Returns:
+            OOASPConfiguration: A configuration with
```
rtaupe: with ...?

asp_interactive_configuration/ooasp/tests/test_metro.py
```diff
@@ -0,0 +1,158 @@
...
+# SPDX-License-Identifier: MIT
+
+from ooasp.interactive import InteractiveConfigurator
+from ooasp.kb import OOASPKnowledgeBase
+from ooasp import settings
+from ooasp.tests.utils import new_iconf
+
+import pytest
+from importlib import reload
+
```
rtaupe, line 11: Several test_metro test cases are currently failing. The issue seems to be related to fclingo.

> FAILED ooasp/tests/test_metro.py::test_metro_next - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_options - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_cv_wagon_constraints - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_cv_wagon_wrong_nr_pass - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_cv_handrail_required - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_multi_wagon - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_cv_wagon - TypeError: object of type 'int' has no len()
> FAILED ooasp/tests/test_metro.py::test_metro_cv_seat_color - TypeError: object of type 'int' has no len()

rtaupe (changes requested):
Thank you for these contributions!
Please just have a look at my comments (mainly concerning typos).

asp_interactive_configuration/DOC.md
susuhahnml: I added a phrase above, corrected the order and rephrased this part.

asp_interactive_configuration/ooasp/config.py
susuhahnml: No, it is the id of the. The object can be either left or right of the association. I just rephrased it to be more clear

asp_interactive_configuration/ooasp/tests/test_metro.py
susuhahnml: Yes, this is an issue with fclingo, since it is still on development, the new version broke things. I will fix the version and push this change.

asp_interactive_configuration/poetry.lock
susuhahnml: This is because we need clingo 5.7 for fclingo to work. This version is only under pypi.test. I am not sure how to add this additional url to the poetry file. I will take a look.

asp_interactive_configuration/DOC.md
susuhahnml: I rephrased, the details and examples are in the subsections below

susuhahnml:
@rtaupe I checked all the comments and added your suggestions. My last commit should have fixed the fclingo issue. Could you try again the poetry install and running the tests?

rtaupe:
This seems to have fixed it, thanks a lot!

rtaupe (approved):

PR #36, opened 2023-11-04
Merge last months' developments to main

rtaupe:


asp_interactive_configuration/examples/metro/fconstraints.lp
```diff
@@ -0,0 +1,79 @@
...
+% Handrail required if standing_room>0
+ooasp_cv(stand_iff_handrail,W,"Handrail required if standing room is larger than 0",(new_object,)):-
+    active(new_object),
+    ooasp_isa(wagon,W),
+    #count{ID2:ooasp_assoc_gen(wagon_handrail,OPT,W,ID2)}=0,
+    &fsum{ooasp_attr_fvalue(standing_room,W)}>0.
+
+
+% the number of associated seats must be equal to the attribute nr_seats
+ooasp_cv(seats_eq__nrseats,W,"Number of seats should be {}, since there are {} Seats associated",(S,new_object)):-
```
gschennersie: The explanation given here is wrong. The second {} will be bound to new_object, which for nr_seats = 10, new_object=1 and no associated seats results in a message like:
"Number of seats should be 0, since there are 1 Seats associated"

gschennersie (changes requested):
Beside the wrong explanation in the metro example (see comment), everything is ok (Test and Notebooks).
One additional minor observation: In the racks example once a Rack and its Frames are created by the solver, removing a link between the Rack and a Frame on the UI has no effect i.e. the association is unchanged. We don't have to fix this before the merge. I suggest testing the example domains after the merge.

gschennersie (approved):

PR #52, opened 2024-07-18
Restructuring the repository (setting asp_interactive_configuration as root)

Kriplingo:
- Created new directory under name *old_encodings* to where the original contents of the project in root directory were moved.
- Moved all contents from *asp_interactive_configuration* directory to the root directory
- Updated `.gitignore` file to fit the new file structure (this includes the new *out* location and the benchmark text files)
- Removed the automatically generated images in *out* directories that were already tracked on git
- Removed `apt.txt` file (with seemingly no use)

> README.md files have been switched accordingly, but changes will still be needed.


rtaupe (approved):
Great, thank you! :-)

PR #55, opened 2024-08-05
Sh/global cautious

susuhahnml:


Kriplingo (commented):
There is a lot of generated output files that were missed by .gitignore, because the saving directory was changed, I suggest I remove them before merging and update .gitignore to catch them. @susuhahnml, @rtaupe , please let me know if I should do this or we should leave this as is.

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
+# Corner cases for benchmarking and inspecting
+
+
+## Lowerbound
+
```
rtaupe, line 5: Why is this section empty?

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+```console
+python ooasp/run.py --frame 9
+```
+
+This case is similar with 17 frames but in that case it gets really slow.
+
+### Smart expansion
+
+1. *(object based)* One frame complains about needing a rack. So it is added
+2. *(global upper)* The single frames don't know a rack is missing because they all think they can use the existing one. So the global constraint counts that if there is 9 frames the upperbound of the rack is filled and there has to be one more. So a second rack is added
```
rtaupe:
```suggestion
2. *(global upper)* The single frames don't know a rack is missing because they all think they can use the existing one. So the global constraint counts that if there is 9 frames the upper bound of the rack is filled and there has to be one more. So a second rack is added
```

benchmarks/README.md
```diff
@@ -0,0 +1,46 @@
...
+3. Nothing else left to infer because both racks think they can fill their lowebound for frames with the existing ones. But this is not true.
```
rtaupe:
```suggestion
3. Nothing else left to infer because both racks think they can fill their lower bound for frames with the existing ones. But this is not true.
```

ooasp/encodings/ooasp_simple.lp
```diff
@@ -19,17 +19,23 @@ check_permanent_cv. % Checks for permanent constraints apply
 #include "ooasp_aux_config.lp". % Auxiliay predicates fot the config
 #include "ooasp_guess.lp". % Guess section
 #include "ooasp_check.lp". % Check section
-#include "ooasp_cautious_opt.lp". % Optimizations used for cautious reasoning
+% #include "ooasp_cautious_opt.lp". % Optimizations used for cautious reasoning
```
rtaupe: Please remove commented-out code entirely, or add a comment explaining why it is only temporarily commented-out. Please do this for all commented-out lines of code.

ooasp/run.py
```diff
@@ -60,49 +72,92 @@ def _get_cautious(ctl, project=False):
     ctl.assign_external(Function("check_potential_cv"), False)
     ctl.configuration.solve.models = "0"
     ctl.configuration.solve.enum_mode = "cautious"
-    ctl.configuration.solve.opt_mode = "optN"
-    if project:
-        ctl.configuration.solve.project = "project"
+    # if project:
```
rtaupe: Please remove commented-out code entirely, or add a comment explaining why it is only temporarily commented-out. Please do this for all commented-out lines of code.

ooasp/run.py
```diff
@@ -231,32 +278,30 @@ def on_model(m):
...
+        ground(ctl, next_id, "object")
+        next_id += 1
 
     end = time.time()
     stats[next_id - 1] = ctl.statistics["summary"]["times"]
 
-    print("Done!")
-    out_name = f"benchmarks/latest/{generate_output_path(args)}"
+    log("Done!")
+    out_name = f"benchmarks/results/second-phase/latest/{generate_output_path(args)}"
```
rtaupe: What is the "second phase"?

rtaupe (changes requested):
Thank you!
Please address my comments.
Also, test_extend_propagate_symmetry is currently failing. Please check if you can reproduce.

ooasp/run.py
susuhahnml: I am not sure, I believe this was @Kriplingo s addition

benchmarks/run.sh
```diff
@@ -10,11 +10,12 @@ file=bench_outputs_$id.txt
 mkdir -p benchmarks/outputs
 echo -e "\e[36mOutputs will be saved to: benchmarks/outputs/$file\e[39m"
 echo -e "\e[35mStarting\e0\e[5m...\e[25m\e[39m"
-for i in {1..20}
+for i in {1..10}
```
rtaupe: If the scope of the benchmarks was changed for performance reasons, please leave a comment here that the whole benchmark set ranges until 20

ooasp/encodings/ooasp_simple.lp
rtaupe: This comment is still open

ooasp/run.py
rtaupe: Since nobody knows what "second phase" refers to, @Kriplingo will revert this.

rtaupe (changes requested):

rtaupe (approved):
Thank you!

PR #58, opened 2024-08-21
Nico/benchmarks

susuhahnml:


susuhahnml:
Hi @Kriplingo! Here is our update. There is a timeout parameter in the command line that we can use so that clingo is properly stoped and the results are still printed. Running the command line (without verbose) will print only the json output with the information. So you can also just save it with `>` By default no smart functions are used. You can add them in a single string: `--smart-generation "object_needed,global_ub,global_lb,association_needed"` The idea would be to vary the order of this functions and the initial objects on the benchmarks. Let us know if this works in your setting and if this gives you interesting results.

Kriplingo:
> Hi @Kriplingo! Here is our update. There is a timeout parameter in the command line that we can use so that clingo is properly stoped and the results are still printed. Running the command line (without verbose) will print only the json output with the information. So you can also just save it with `>` By default no smart functions are used. You can add them in a single string: `--smart-generation "object_needed,global_ub,global_lb,association_needed"` The idea would be to vary the order of this functions and the initial objects on the benchmarks. Let us know if this works in your setting and if this gives you interesting results.

@susuhahnml these are great changes, but I have ran into a bug, where if the `--timeout` argument is not provided, the program crashes with the following error message:
```
Traceback (most recent call last):
  File "C:\Users\z004zdbc\Desktop\fresh\OOASP\ooasp\run.py", line 566, in <module>
    solver.run()
  File "C:\Users\z004zdbc\Desktop\fresh\OOASP\ooasp\run.py", line 521, in run
    time_left = self.timeout - (time.time() - run_start)
                ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```
I am quite certain this is unwanted behaviour, and we should either set a default (perhaps an infinity constant), or change the code to avoid this situation. Please let me know if I should attempt a fix.

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
+            self.times["smart_generation"]["time"] += time.time() - start
+            if things_done:
+                continue
+            self.log(subtitle(f"Solving for size {self.next_id - 1}...", "RED"))
+            self.ctl.configuration.solve.models = "1"
+            start = time.time()
+            with self.ctl.solve(
+                assumptions=self.assumptions, async_=True, on_model=on_model
+            ) as hdl:
+                time_left = self.timeout - (time.time() - run_start)
```
Kriplingo, line 521: This line causes a crash if timeout is not specified in the program arguments.

Kriplingo (changes requested):

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
+            things_done = self.smart_generation()
+            self.times["smart_generation"]["time"] += time.time() - start
+            if things_done:
+                continue
+            self.log(subtitle(f"Solving for size {self.next_id - 1}...", "RED"))
+            self.ctl.configuration.solve.models = "1"
+            start = time.time()
+            with self.ctl.solve(
+                assumptions=self.assumptions, async_=True, on_model=on_model
+            ) as hdl:
```
Kriplingo:
```suggestion
            ) as hdl:
            self.timeout = float('inf') if self.timeout is None else self.timeout
```

Kriplingo: This change should avoid cases, where `--timeout` option not being specified causes a crash, but this change could also be made to the default `OOASPRacksSolver.timeout` value.

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
-    parser.add_argument("--moduleII", type=int, default=0)
-    parser.add_argument("--moduleIII", type=int, default=0)
-    parser.add_argument("--moduleIV", type=int, default=0)
-    parser.add_argument("--moduleV", type=int, default=0)
-    return parser
+        return added > 0
+
+    def global_ub(self) -> bool:
+        """
+        The appearance of predicate global_ub(C2, X, new_object)
```
nrueh:
```suggestion
        The appearance of predicate global_ub(C2, N, new_object)
```

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
+                self.log("\t  ---> Apply ", s)
+                c2, needed, _ = s.arguments
+                for _ in range(0, needed.number):
+                    self.add_object(c2.name)
+                return True
+        return False
+
+    def global_lb(self) -> bool:
+        """
+        The appearance of predicate global_lb(C1, X, new_object)
```
nrueh:
```suggestion
        The appearance of predicate global_lb(C1, N, new_object)
```

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
+        return False
+
+    def association_needed(self) -> bool:
+        """
+        The appearance of predicate association_needed(ASSOC, ID1, ID2, new_object)
+        in the brave consequences indicates the need to add an association between two objects.
+
+        We know that ID1 needs at least one object of class C1 and ID2 needs at least one object of class C2
+        We know that the objects ID1 and ID2 can potentially be associated by ASSOC
+        We also know that the classes of these objects was set byt the user or the smart association.
```
nrueh:
```suggestion
        We also know that the classes of these objects was set by the user or the smart association.
```

ooasp/run.py
```diff
@@ -1,314 +1,566 @@
...
+        return False
+
+    def global_lb(self) -> bool:
+        """
+        The appearance of predicate global_lb(C1, X, new_object)
+        in the cautious consequences indicates the need to add N objects of type C1
+
+        Given a target association ASSOC where each C2 can be associated to at least MIN objects of C1
+        and each C1 has to be associated to exactly one C2.
+        We count the global number of objects of class C1 and this is not enough to cover all C2 lowe bounds.
```
nrueh:
```suggestion
        We count the global number of objects of class C1 and this is not enough to cover all C2 lower bounds.
```

ooasp/encodings/ooasp_smart_generation.lp
```diff
@@ -0,0 +1,115 @@
...
+    % The amount of elements C1 does not fill the lowe bound of C2
+    N2 * MIN > N1,
+	active(new_object).
+
+% -----------------------------------
+% association_needed(ASSOC, ID1, ID2, new_object).
+%   The association ASSOC bewtwen objects ID1 and ID2 is needed
+%   We know that ID1 needs at least one object of class C1 and ID2 needs at least one object of class C2
+%   We know that the objects ID1 and ID2 can potentially be associated by ASSOC
+%   We also know that the classes of these objects was set byt the user or the smart association.
```
nrueh:
```suggestion
%   We also know that the classes of these objects was set by the user or the smart association.
```

ooasp/encodings/ooasp_smart_generation.lp
```diff
@@ -0,0 +1,115 @@
...
+    #count { ID:ooasp_isa(C2,ID) } = NUM2,
+    % The amount of elements C2 has already reached the maximum of C1 it can have
+    NUM2 * MAX < NUM1,
+	active(new_object).
+
+% -----------------------------------
+% global_lb(C1,N,new_object).
+%   Given a target association ASSOC where each C2 can be associated to at least MIN objects of C1
+%   and each C1 has to be associated to exactly one C2.
+%   We count the global number of objects of class C1 and this is not enough to cover all C2 lowe bounds.
```
nrueh:
```suggestion
%   We count the global number of objects of class C1 and this is not enough to cover all C2 lower bounds.
```

ooasp/encodings/ooasp_smart_generation.lp
```diff
@@ -0,0 +1,115 @@
...
+%   We count the global number of objects of class C2 and this is not enough to cover all C2 objects
+%   therefore the upper bound of classes C2 was already reached and we need to add N objects of class C2
+%   to fill the gap.
+%     - Argument new_object is added to the predicate to avoid the redefiniton in following states
+%
+%  The appearance of this predicate in the cautious consequences indicates the need to add N objects of type C2
+
+%  EXAMPLE:
+%   We can see the use of this predicate in an example with 9 frames and 1 rack.
+%   Where the need for one more rack since the upperbounf of rack is 8 and the number of frames is 9.
```
nrueh:
```suggestion
%   Where the need for one more rack since the upperbound of rack is 8 and the number of frames is 9.
```

Kriplingo:
@susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?

nrueh:
> @susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?

Hm, that is weird. I ran your command and the first time it worked but now it also does not seem to terminate. Let me have a look

ooasp/run.py
nrueh: I fixed it by adding a default value to `OOASPRacksSolver.timeout`

nrueh:
> @susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?

I fixed this, too, I think by breaking out of the loop before any solving is started in case the timeout is already reached. Currently, the timeout also does not affect the smart generation. I will think about making it more general.

Kriplingo:
> > @susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?
> 
> I fixed this, too, I think by breaking out of the loop before any solving is started in case the timeout is already reached. Currently, the timeout also does not affect the smart generation. I will think about making it more general.

@nrueh One more question, as I am currently testing it further. 
Should the timeout apply to grounding as well? Because currently even in the new version, it does not seem to affect grounding: `python ooasp/run.py --elementA 20 --elementB 20 --elementC 20 --elementD 20 --timeout 1` does not seem to terminate on timeout and gives this result (which does not seem correct if it should apply to grounding, and if it weren't to, I don't quite understand why solving would not even start):
```
{
    "timeout": true,
    "#objects": 80,
    "#objects_added_per_type": {
        "elementA": 20,
        "elementB": 20,
        "elementC": 20,
        "elementD": 20
    },
    "#associations": 0,
    "times": {
        "runtime": 41.895,
        "initialization": 41.895,
        "smart_generation": {
            "time": 0.0,
            "cautious": 0,
            "brave": 0,
            "functions": {}
        },
        "solve": 0,
        "ground": 41.891
    },
    "model": null
}
```

nrueh:
> > > @susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?
> > 
> > 
> > I fixed this, too, I think by breaking out of the loop before any solving is started in case the timeout is already reached. Currently, the timeout also does not affect the smart generation. I will think about making it more general.
> 
> @nrueh One more question, as I am currently testing it further. Should the timeout apply to grounding as well? Because currently even in the new version, it does not seem to affect grounding: `python ooasp/run.py --elementA 20 --elementB 20 --elementC 20 --elementD 20 --timeout 1` does not seem to terminate on timeout and gives this result (which seems okay if timeout should not apply to grounding):
> 
> ```
> {
>     "timeout": true,
>     "#objects": 80,
>     "#objects_added_per_type": {
>         "elementA": 20,
>         "elementB": 20,
>         "elementC": 20,
>         "elementD": 20
>     },
>     "#associations": 0,
>     "times": {
>         "runtime": 41.895,
>         "initialization": 41.895,
>         "smart_generation": {
>             "time": 0.0,
>             "cautious": 0,
>             "brave": 0,
>             "functions": {}
>         },
>         "solve": 0,
>         "ground": 41.891
>     },
>     "model": null
> }
> ```
Currently it does not affect grounding, but in my opinion it should be applied globally (so grounding and everything else). Tt was tricky implementing it like that. But we are thinking about ways to improve it. In case you have any ideas, go ahead :)

 Also we would like to make it possible to kill it from some external tool (e.g. runsolver) and still get the stats but this was also not easy...

Kriplingo:
> > > > @susuhahnml @nrueh also I don't think the timeouts are working properly, either I am calling it wrong, or there is some mistake when it comes to terminating the solving early. `python ooasp/run.py --elementA 4 --elementB 4 --elementC 4 --elementD 4 --timeout 1 --smart-generation "association_needed"` command has been running on my computer for about 20 minutes now. Am I calling it wrong?
> > > 
> > > 
> > > I fixed this, too, I think by breaking out of the loop before any solving is started in case the timeout is already reached. Currently, the timeout also does not affect the smart generation. I will think about making it more general.
> > 
> > 
> > @nrueh One more question, as I am currently testing it further. Should the timeout apply to grounding as well? Because currently even in the new version, it does not seem to affect grounding: `python ooasp/run.py --elementA 20 --elementB 20 --elementC 20 --elementD 20 --timeout 1` does not seem to terminate on timeout and gives this result (which seems okay if timeout should not apply to grounding):
> > ```
> > {
> >     "timeout": true,
> >     "#objects": 80,
> >     "#objects_added_per_type": {
> >         "elementA": 20,
> >         "elementB": 20,
> >         "elementC": 20,
> >         "elementD": 20
> >     },
> >     "#associations": 0,
> >     "times": {
> >         "runtime": 41.895,
> >         "initialization": 41.895,
> >         "smart_generation": {
> >             "time": 0.0,
> >             "cautious": 0,
> >             "brave": 0,
> >             "functions": {}
> >         },
> >         "solve": 0,
> >         "ground": 41.891
> >     },
> >     "model": null
> > }
> > ```
> 
> Currently it does not affect grounding, but in my opinion it should be applied globally (so grounding and everything else). Tt was tricky implementing it like that. But we are thinking about ways to improve it. In case you have any ideas, go ahead :)
> 
> Also we would like to make it possible to kill it from some external tool (e.g. runsolver) and still get the stats but this was also not easy...

Yes, makes sense, python is not a nice tool to work with when it comes to timeouts, signals and multi-threading in general. I will try to think a bit on it, but currently I don't really know if there is some other way to make sure we get the output even if the process is forcefully killed, other than maybe keeping track of the state periodically and saving it in a variable which would be printed out on killing of the process, or saved in a file, but I am quite sure this would come at a cost of performance. (Not to mention the fact that I am not sure if this would even be possible with the clingo API control).

nrueh:
Sorry, I accidentally pushed something I was trying out with heuristics in this branch but it is not supposed to go in here...I didn't have time to fix it now but I will revert it tomorrow!

nrueh:
We could also not care about the output for benchmarking...in the end we are interested in seeing what options work and what kind of instances we can solve, right?

nrueh:
> Sorry, I accidentally pushed something I was trying out with heuristics in this branch but it is not supposed to go in here...I didn't have time to fix it now but I will revert it tomorrow!

This is fixed now.

rtaupe:
@Kriplingo Thank you for your in-depth reviewing and commenting! :-)

rtaupe:
> We could also not care about the output for benchmarking...in the end we are interested in seeing what options work and what kind of instances we can solve, right?

I care about the stats if solving terminates. If it doesn't, I think they are less important.

ooasp/encodings/ooasp_simple.lp
```diff
@@ -18,13 +18,8 @@ check_permanent_cv. % Checks for permanent constraints apply
...
+#include "ooasp_smart_generation.lp". % Functions for smart generation
 
-#show lb_at_least/6.
-#show upper_filled/5.
-#show lower_global/5.
+#show .
 
-#program domain(new_object, cls).
-
-:- user(ooasp_isa(C,new_object  )),
```
rtaupe, line 29: Apparently this was removed, not moved.
Why?

ooasp/encodings/ooasp_smart_generation.lp
```diff
@@ -0,0 +1,115 @@
...
+
+
+#show object_needed/6.
+#show global_ub/3.
+#show global_lb/3.
+#show association_needed/4.
+
+% Target associations are those with cardinality 1..1-MIN..MAX
+% We only consider this type of associations for the bound calculations
+is_target_assoc(ASSOC,OPT):-
```
rtaupe: I know more or less what `OPT` is referring to, but why is it called OPT?
Please document the parameter here, and also where it is originally defined (I think this is in `ooasp_aux_kb.lp`)

ooasp/encodings/viz_config.lp
```diff
@@ -12,14 +12,15 @@ attr(graph, config, splines, ortho).
 attr(graph, config, nodesep, "1").
 attr(graph, config, rankdir, "TB").
 attr(graph, config, size, "40,100").
+attr(graph, config, (label,0),T):-title(T).
```
rtaupe: Where is the `title` predicate defined?

ooasp/run.py
```diff
@@ -1,314 +1,591 @@
...
+                                           is not None else [])
+        self.view = view
+        self.timeout = timeout if timeout is not None else 1200
+
+        self.ctl = Control([
+            "1",
+            "--warn=none",
+            "-c config_name=c1",
+            "-c kb_name=k1",
+            "-t3",
```
rtaupe: What's `t3`?

rtaupe (changes requested):
I haven't yet finished reviewing this PR, but a few questions already came up. Please address them, thanks!

ooasp/run.py
nrueh: Instruction for clingo to use 3 threads, the verbose command would be `--parallel-mode=3`. Do you prefer that?

ooasp/encodings/ooasp_smart_generation.lp
nrueh: I would be in favor of renaming the `OPT` variable to something more descriptive. It stands for "option" and specifies the side of the association (1 or 2).

ooasp/run.py
rtaupe: Yes please. I can't find `t3` in `clingo --help=3`. Please also add brief explanation as a comment on that line. Thanks!

nrueh:
```suggestion
            "--parallel-mode=3", # Run parallel search with 3 threads
```

ooasp/encodings/ooasp_simple.lp
nrueh: @susuhahnml Do you know why? Was this only temporary or are we not using the `user(...)` predicates anymore?

ooasp/encodings/viz_config.lp
nrueh: I removed this line. Seems to be old code

ooasp/encodings/ooasp_simple.lp
susuhahnml: Yes, at the moment we are adding the user predicate as atoms, not assumptions so this was not really making any difference since it always the case bacause of how we add things. We will add this properly when we include interactivity.

ooasp/encodings/ooasp_smart_generation.lp
susuhahnml: Done

Kriplingo (dismissed):
Code works on my side, I just need to know one more thing. Did you remove the option for choosing a number of threads (or renamed it)? So that I know if I need to take the number of threads into account for the benchmarks?

nrueh:
> Code works on my side, I just need to know one more thing. Did you remove the option for choosing a number of threads (or renamed it)? So that I know if I need to take the number of threads into account for the benchmarks?

Yes, but now you can use all normal clingo options on the command line when running the program.
The corresponding option in clingo is `--parallel-mode` or just `-t`

So for running with 3 threads for example:
```
python ooasp/app.py --frame 5 --parallel-mode=3
```

nrueh:
We also removed the timeout option. If I understood well you were enforcing this with some external tool, anyways (which I  think is the best way to go).
However, in case you need it clingo also has a built-in timeout option `--time-limit`

Kriplingo:
> We also removed the timeout option. If I understood well you were enforcing this with some external tool, anyways (which I think is the best way to go). However, in case you need it clingo also has a built-in timeout option `--time-limit`

Yes, I use an external tool to enforce it, but just out of curiosity, since it is in-built by clingo, does this apply to grounding as well?
Also thank you for all the changes, they have been extremely helpful.

nrueh:
> > We also removed the timeout option. If I understood well you were enforcing this with some external tool, anyways (which I think is the best way to go). However, in case you need it clingo also has a built-in timeout option `--time-limit`
> 
> Yes, I use an external tool to enforce it, but just out of curiosity, since it is in-built by clingo, does this apply to grounding as well? Also thank you for all the changes, they have been extremely helpful.

Yes, it applies to everything we are doing inside the Application class. So that also involves grounding, adding objects, (cautious/brave) solving, etc...

ooasp/encodings/ooasp_smart_generation.lp
```diff
@@ -0,0 +1,110 @@
...
+is_target_assoc(ASSOC,SIDE):-
+    ooasp_assoc_limit(ASSOC,max,SIDE,C1,1,C2),
+    ooasp_assoc_limit(ASSOC,min,SIDE,C1,1,C2).
+
+% =====================
+% Incremental generation
+% =====================
+#program domain(new_object, cls).
+
+% #heuristic object_needed(ID1, ASSOC, A, C2, SIDE, new_object):object_needed(ID1, ASSOC, A, C2, SIDE, new_object),ooasp_assoc_specialization(ASSOC,_). [1,sign]
```
rtaupe: Please don't include commented-out code; or at least add an explanation why it is commented-out.

rtaupe (changes requested):
Thanks again -- almost ready now :)
Just explain the one line of commented-out code, please.

ooasp/encodings/ooasp_smart_generation.lp
susuhahnml: Just removed the line.

rtaupe (approved):
Thank you!

PR #60, opened 2024-09-26
Object bounds

susuhahnml:


ooasp_configuration.lp
rtaupe, line 1: What is this file why has it changed?
Should it really be under version control?

ooasp_configuration_wrong.lp
rtaupe, line 1: What is this file why has it changed?
Should it really be under version control?

ooasp/encodings/ooasp_guess.lp
```diff
@@ -1,6 +1,17 @@
 % Copyright (c) 2022 Siemens AG Oesterreich
 % SPDX-License-Identifier: MIT
 
+#program include(new_object, cls).
+
+% Include for performance
+ooasp_isa(cls,new_object).
+
+% Fact used in the UI to indentify objects that must be included
```
rtaupe:
```suggestion
% Fact used in the UI to identify objects that must be included
```

rtaupe (changes requested):
Thank you!
I just have a few questions to address, please.

ooasp_configuration.lp
susuhahnml: This two files were generated output. I just removed them

rtaupe (approved):
Thank you!

PR #61, opened 2024-10-03
Implement feedback from Richard

nrueh:


nrueh:
@rtaupe I implemented the changes you suggested. Please have a look!

Unfortunately, there also seem to be some formatting changes...I have the pre-commit installed so I am not sure if this these changes are due to bad formatting before or me changing the formatting...

rtaupe:
@Kriplingo could you please investigate whether the new version of these files conforms to the formatting style as intended?

Kriplingo:
> @Kriplingo could you please investigate whether the new version of these files conforms to the formatting style as intended?

According to the autopep, they should be.

Kriplingo:
@nrueh @rtaupe  Can I have one comment to this PR? Since the branch this PR is based on, was branched from main and not from the clinguin branch (which was the newest one for quite some time and I assume we did not manage to merge it because of all the hurry around the extended abstract) we are bound to run into merge conflicts sooner or later. Would it be possible to merge this branch into the clinguin one and then merge only that request into main, so that I can resolve the conflicts as soon as possible? This is purely because I do not have the right to merge into main so we would potentially avoid "one extra round of code reviews" and I can resolve the conflicts faster.

rtaupe (dismissed):
Thank you!

rtaupe (approved):

PR #68, opened 2024-11-12
removed quotations from boolean definition

### LINKED ISSUE #67: Inconsistent definition of attribute type "boolean"
Currently the type for bool is defined as `"boolean"` while other types do not include quotation marks, making definition of own domains confusing and inconsistent.

Kriplingo:
I removed the quotation marks, as they do not seem to break anything.


rtaupe (approved):

susuhahnml:
There is no example for instances using this type of attribute so no change was needed anywhere else. Would be nice to have an example

